Issue #4, opened 2014-12-08
Bundle KA Lite into PyRun


aronasorman:
<a href="https://github.com/aronasorman"><img src="https://avatars.githubusercontent.com/u/191955?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [aronasorman](https://github.com/aronasorman)**
_Thursday Dec 04, 2014 at 19:11 GMT_
_Originally opened as https://github.com/learningequality/ka-lite/issues/2761_

----

[PyRun](https://www.egenix.com/products/python/PyRun/) is the most promising way right now for us to bundle Python to Unix systems. We should see if we can do it for KA Lite and see if it runs. 

If it does work, let's create a script that outputs KA Lite + PyRun. We should also determine what settings we need to change in order for it to work perfectly. These changes would then go to the `local_settings.py` inserted by the build server.

aronasorman:
@cpauya can you give an update on where you are with testing out PyRun?

cpauya:
We can use PyRun as a `virtualenv` replacement as I've tested earlier.

But since packaging KA-Lite into PyRun under OSX 10.10.1 doesn't work well (compiler issues which will take a while) - what I will do now is to create a script to package them together and make sure that KA-Lite uses PyRun (the Python version of PyRun) instead of the system Python.

PR #9, opened 2015-01-08
Initial script to install pyrun and bundle ka-lite inside.

cpauya:
Hi @aronasorman - this is the very initial script for #4 

Done:
* It downloads the official PyRun installer script and sets-up a minimal PyRun directory with Python 2.7.8 (more than 25mb) under `/tmp/ka-lite-on-pyrun/`.

ToDos:
* git clone ka-lite on the PyRun tmp folder
* modify ka-lite`s `/scripts/python.sh` script to point to PyRun`s `/bin/pyrun` folder so it uses the Python version of PyRun instead of the system Python.

### DIFF
--- a/osx/ka-lite-on-pyrun.sh
+++ b/osx/ka-lite-on-pyrun.sh
@@ -10,18 +10,45 @@
 # . Clone KA-Lite into a directory
 # . Create a kalite binary in `/usr/local/bin/` to use PyRun's Python instead of the system Python.
 # . Modify `<ka-lite-folder>/python.sh` so it uses PyRun`s Python.
+#
+# TODO:
+# * use `tempfile.py` instead of mktemp which is "subject to race conditions"
 
-INSTALL_PYRUN="install-pyrun.sh"
-PYRUN_DIR="/tmp/ka-lite-on-pyrun"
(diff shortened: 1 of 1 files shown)


osx/ka-lite-on-pyrun.sh
```diff
@@ -0,0 +1,27 @@
...
+# . git
+#
+# Steps
+# . Execute the PyRun installer from https://downloads.egenix.com/python/install-pyrun into a tmp directory.
+# . Clone KA-Lite into a directory
+# . Create a kalite binary in `/usr/local/bin/` to use PyRun's Python instead of the system Python.
+# . Modify `<ka-lite-folder>/python.sh` so it uses PyRun`s Python.
+
+INSTALL_PYRUN="install-pyrun.sh"
+PYRUN_DIR="/tmp/ka-lite-on-pyrun"
```
aronasorman: I wonder if there's a way to get a platform's tmp dir?

aronasorman: I guess `mktemp` is our friend here: http://unixhelp.ed.ac.uk/CGI/man-cgi?mktemp

aronasorman:
Let's get this in now and then fix when we're doing the OS X installer.

Issue #53, opened 2015-06-01
Install pip dependencies when running ka-lite windows installer


amodia:
**Branch**: Master
**Target**: 0.14.x
**Expected behavior**: When running `ka-lite windows installer` it must automatically install the pip dependencies for ka-lite.
**Actual behavior**: Pip dependences is not automatically install when running `ka-lite windows installer` this issue is related to #48 .

cpauya:
Adding the instructions from @aronasorman here:
...
[May-27 10:43 PM] Aron Fyodor Asor: yeah, so basically the windows installer isn't working now, since it doesn't grab the dependencies from pip
the way forward right now is to generate a static zip package with all dependencies by running 'python setup.py sdist --static'
that would create a zip file inside ka-lite/dist
now that is what will be included in the windows installer
and once installation on windows starts, you should do a 'pip install <zipfile>' to install.
...

PR #57, opened 2015-06-10
Use kalite command instead of deprecated scripts

### LINKED ISSUE #56: Windows task tray program does nothing
I built the installer as per instructions using the `develop` branch of ka-lite and tried installing on Windows 7. The task tray icon is unable to start the server; clicking it does not cause the "Start server" option to become disabled (which is what I expected). `kalite start` works fine. @ruimalheiro @cpauya I'm assigning myself, but if someone more savvy decides to tackle this I would not object.

MCGallaspy:
Fixes #56 (or at least it ought to). Going to build the gui and give it a whirl, but @cpauya can you try it out, too?


MCGallaspy:
Still testing this out. Reviewer, please build the installer and see if the systray program functions.

MCGallaspy:
Oof, the size of the executable exploded. I wonder if some Windows wizard can build it smaller?

MCGallaspy:
This appears to be the right approach (the sys tray now works), but starting the server that way seems to be missing all the static files (like css assets).

MCGallaspy:
@amodia if you're working on this as well, take note ^

amodia:
Ok @MCGallaspy I will stop working on it I will test this if works fine on me.

amodia:
@MCGallaspy I tested this but its not working I use `make.vbs` to build `kalitesetup.exe`. I try also to manually run by using `kalite start` on the command line and it will work fine.  

This is the screenshot that ka-lite server is not starting.
screen shot 2015-06-11 at 1 12 07 pm

cpauya:
Sorry @MCGallaspy - I haven't tested this yet as I'm freeing-up space for VS2015 on my laptop's virtual box.  Solution is not compiling on my existing VS2012 so I'm forced to upgrade, but as per @amodia's testing above, it doesn't seem to work.

Hope by tomorrow I could test this one out unless someone else beats me to it.

MCGallaspy:
@cpauya try checking out the solution file from master and compiling it
that way with VS 2012 if it's easier for you. I only opted for 2015 because
the price was right.

cpauya:
Thanks @MCGallaspy.

Ok, was able to build it using VS2012-with-update4 by setting the `Platform Toolset` to `Visual Studio 2012 - Windows XP (v110_xp)` - see screenshot below.  It was complaining about using the `v140_xp` build update before (perhaps because it was expecting me to be using VS2015):
ie11-win7-kalite-2

Then after building it, I have to copy the `ka-lite` repo to the `gui-packed` folder to make the built `KA Lite.exe` systray icon work.  Please note that I have installed `python-setup/python-2.7.10.msi` earlier - I'm just being lazy here instead of building the whole installer and installing it - but it should be working the same.

Here's a .gif of it working with start/stop commands using `bin/kalite`:
ka-lite
 
So yes, this PR works. :+1:

MCGallaspy:
@cpauya great, thanks! When you start the server using the sys tray program, does it fail to load static assets? @amodia what OS are you using?

cpauya:
No, it actually loaded the whole site and asked me to create a superuser account, it's shown somewhere on timeframe **1:20** of the .gif above.  

I will double-check with @amodia tomorrow (by creating an installer and using that) just to be sure coz he fixed the #53 regarding loading static assets with `pip`. :smile_cat:

MCGallaspy:
Okay, I'll build a smaller installer and then let's test this out on various flavors of Windows.

Works ok on:
- [ ] Windows XP (using any browser other than IE8)
- [ ] Windows 7
- [ ] Windows 8

amodia:
Hi @MCGallaspy I'm using `Windows 7 professional N  32-bit` and still not working on me. it's just showing server starting but not running the `ka-lite server`.

ka-lite-a

MCGallaspy:
@amodia it may just be very slow.

MCGallaspy:
Okay, I'm having a similar issue on Windows 8. Testing out `kalite start` I get the following error:
```
Error in migration: updates:0004_auto__add_field_updateprogresslog_current_stage

DatabaseError: table "_south_new_updates_updateprogresslog" already exists
```
I think it must be an issue with the install process, not the GUI.

amodia:
@MCGallaspy I also test this on `windows xp sp3` and I got `Unknown Command "migrate"` error when I manually run the `kalite start` on `program files\ka-lite` directory. I think the `south package` not currently in `site-packages`. 

screen shot 2015-06-15 at 3 48 49 pm


This is the behavior when I manually run the `kalite start command` on `Installers\windows\ka-lite` directory using command line refer to the `.gif` below. 
ka-lite-a

MCGallaspy:
If anything, it seems like a missing module was not an error with the gui but rather during the installer build process. So I'll merge this one and we can sort out the rest later. @amodia can you open new issues regarding what you wrote above?

Issue #69, opened 2015-07-01
KaliteSetupScript.iss syntax error


MCGallaspy:
On line 345 on `master`. Encountered when I tried to build the installer.

MCGallaspy:
Column 5. No character is visible there on github, but in my editor it's definitely a weird unicode character.

PR #70, opened 2015-07-01
Fix up KaliteSetupScript.iss

MCGallaspy:
An issue with the encoding, presumably...?

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -451,8 +451,8 @@ begin
         if installFlag then
         begin
             HandlePipSetup();
-            setupCommand := 'kalite manage setup --noinput -o "'+ServerInformationPage.Values[0]+'" -d "'+ServerInformationPage.Values[1]+'" -u "'+UserInformationPage.Values[0]+'" -p "'+UserInformationPage.Values[1]+'"';
-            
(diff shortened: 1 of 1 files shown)


windows/installer-source/KaliteSetupScript.iss
```diff
@@ -342,7 +342,7 @@ var
 
 begin
     PipPath := 'C:\Python27\Scripts\pip.exe';
-    ﻿PipCommand := 'install ' + '"' + ExpandConstant('{app}\ka-lite\dist\ka-lite-static-')  + '{#MyVersion}' + '.zip' + '"';
+    PipCommand := 'install ' + '"' + ExpandConstant('{app}\ka-lite\dist\ka-lite-static-')  + '{#MyVersion}' + '.zip' + '"';
```
MCGallaspy: Also, `{#MyVersion}` returns a different version string than the output of `setup.py sdist --static` (namely `0.14alpha1` vs `0.14a1`). Let me try to resolve that in this PR as well.

MCGallaspy:
Fixes #69.

MCGallaspy:
Fixes #71 also.

aronasorman:
Have you tested the changes in here?

MCGallaspy:
I've tested everything except the latest commit. It's exhibiting different behavior from earlier 0.14.x builds -- previously the db setup failed almost instantaneously, and now it's taking a loooooong time. But that's consistent with versions, so I'm taking it as a good sign.

MCGallaspy:
Yeah, I think that step is hanging because it's waiting for input or something silly. Will amend again.

MCGallaspy:
Tested and it works.

Issue #72, opened 2015-07-01
Windows -- ensure there is a user data migration path for 0.13.x to 0.14.x


MCGallaspy:
Now that many user data files are unbundled from the source directory.

benjaoming:
@MCGallaspy wasn't the final verdict that *there is no migration path* ?

MCGallaspy:
Dunno what you mean by "there is no migration path". Fixed in #73.

PR #73, opened 2015-07-02
Windows migration path

MCGallaspy:
A work in progress. Starting by refactoring the code for readability. Up to https://github.com/learningequality/installers/commit/78541faeb3b0cf892dbc2d4210d366617ae2018e it should be equivalent. TODO: In the event that an old installation is detected and the user wishes to upgrade (as opposed to trampling old data), then provide a migration mechanism by calling the "gitmigrate" management command after new version is installed. Ensure that removing the old installation files doesn't trample the old user data.


MCGallaspy:
Quite a few changes here. I tried to refactor the `KaliteSetupScript.iss` file in general to make it more readable (mostly by spinning off long code sections into procedures and using semantic variable names instead of long string literals). Additionally:
* Previous version is got from the registry values that InnoSetup (IS) writes for uninstall.
* There's a new migration path for 0.13 (or lower) to 0.14 (or higher) version of KA Lite, as determined by the IS registry values. It goes like this:
  *  If the user chooses to upgrade, then the user's 0.13 database file and content directory will be copied into the home directory of the installing user (`%USERPROFILE%\.kalite\`). This should exactly mimic the gitmigrate command -- it could only fail if the user could somehow specify a non-default CONTENT_ROOT setting, which should be impossible using the Windows installer.
  * If the user does not choose to upgrade, a new database file is created normally (in the same location).

MCGallaspy:
I have an installer (without assessment items) to test the migration path. Will upload and post a link here.

MCGallaspy:
Fixes #72.

PR #87, opened 2015-07-08
Bundle blank db in windows installer

### LINKED ISSUE #80: Bundle pre-migrated empty database into installers to reduce load time
The first time the server is started, it takes a long time to load (see https://github.com/learningequality/installers/issues/78), and the longest-running task is the database migration. When the setup was run from the command line, at least the user saw that something was happening, but now it just seems like it's hanging.

If we can bundle a pre-migrated (but empty) database in (like we do in the Android app), this should cut down the load time very significantly. Steps to create such a database:

- Delete any existing database
- Run `bin/kalite manage syncdb` and then `bin/kalite manage migrate --merge`
- Do *not* run `setup`, or start the server (as that will start adding data)

MCGallaspy:
Add instructions for creating a blank db to be bundled in. Handles copying the blank db to user's data location during install. Fixes #80. Depends on #73.


MCGallaspy:
Will test on Windows 8.

MCGallaspy:
@jamalex works fine on windows 8.

PR #88, opened 2015-07-08
Windows installer - don't assume drive letter for `pip.exe`

### LINKED ISSUE #74: Pip dependencies have failed to install
**Branch**: 0.14 (installing on Windows 7)

**Current Behavior**: Error while installing 0.14 over 0.13 (fully functional with several coaches & learners & language packs):

Pip dependencies have failed to install.

pip-dependencies

(The spelling error has been fixed in https://github.com/learningequality/installers/pull/73)

The installation continued after this error, but is probably why the 0.14 server was unable to start at the end. I had no issues with Pip this morning when I did a clean new install on another Win 7 machine.  

**Steps to reproduce**:

1. Download & install the 0.14 on the machine with previous 0.13 installation of KA Lite

### LINKED ISSUE #81: UX Issue: "Setup will now install pip dependencies"
I clicked out before I could take a screenshot, but this was a popup that appeared after installing. Thought this line might be confusing for people who don't understand development. Could this line be changed to something easier to understand, like "configuring your installation" or something similar?

MCGallaspy:
Handles the case where `pip.exe` is not on the `C:` drive. Fixes #74 and fixes #81. Will test out on Windows 8. @radinamatic I'll upload an installer for you to try, as well.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -417,6 +417,38 @@ begin
     end;
 end;
 
+{ Used in GetPipPath below }
+const
+    DEFAULT_PATH = '\Python27\Scripts\pip.exe';
+
+{ Returns the path of pip.exe on the system. }
+{ Tries several different locations before prompting user. }
+function GetPipPath: string;
+var
+    path : string;
+    i : integer;
+begin
+    for i := Ord('C') to Ord('H') do
+    begin
+        path := Chr(i) + ':' + DEFAULT_PATH;
+        if FileExists(path) then
+        begin
+            Result := path;
+            exit;
+        end;
+    end;
+    MsgBox('Could not find pip.exe. Please select the location of pip.exe to continue installation.', mbInformation, MB_OK);
+    if GetOpenFileName('Please select pip.exe', path, '', 'All files (*.*)|*.*', 'exe') then
+    begin
+        Result := path;
+    end
+    else begin
+        MsgBox('Fatal error'#13#13'Please install pip and try again.', mbError, MB_OK);
+        forceCancel := True;
+        Result := '';
+    end;
+end;
+
 procedure HandlePipSetup;
 var
     PipCommand: string;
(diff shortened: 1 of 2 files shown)


MCGallaspy:
Depends on #73.

MCGallaspy:
Ugh, no, this doesn't seem to work.

MCGallaspy:
Here's my solution:
* The installer searches for pip in `XX`:\Python27\Scripts\pip.exe, where `XX` stands for the letters "C" through "H".
* If it can't find pip in those locations, it prompts the user to select pip.exe.
* If the user cancels the prompt, the installation is aborted.
* If pip is found using either of the above methods, then installation proceeds normally.

radinamatic:
I just now realized that there are 3 **pips** in my `D:\Python27\Scripts` folder, all with exactly the same `Date modified` attribute:
- pip.exe
- pip2.exe
- pip2.7.exe

So, who's the winner?

MCGallaspy:
They all are.

radinamatic:
Meaning, if the user is prompted, they can choose any of those three indifferently?

radinamatic:
Sorry, 32bit... :(

runtime64

MCGallaspy:
> Meaning, if the user is prompted, they can choose any of those three indifferently?

Indeed so. It makes more sense on *nix systems, where they are generally symlinks to the versioned name.

MCGallaspy:
Regarding your other error, I'm on it.

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -417,6 +417,38 @@ begin
...
+    DEFAULT_PATH = '\Python27\Scripts\pip.exe';
+
+{ Returns the path of pip.exe on the system. }
+{ Tries several different locations before prompting user. }
+function GetPipPath: string;
+var
+    path : string;
+    i : integer;
+begin
+    for i := Ord('C') to Ord('H') do
```
aronasorman: Why not until 'Z:\'? I believe Z: is the default drive letter for wine.

MCGallaspy: Who is running this in wine?

MCGallaspy: But at any rate such a user will be prompted to select the location of pip.exe later on.

PR #97, opened 2015-07-14
Require python 2.7.9 on Windows

### LINKED ISSUE #95: KA-Lite installer upgrade from 0.12 to 0.14 fails to update Python version
Windows XP most but other WIndows version may be affected:

While testing the 12.11 to 14 installer now under development I discovered that Python was not being updated to the 14 bundled version (Python 2.7.10) and pip is needed yet not available in the (2.7.3 Python) 12.11 bundled version. To replicate install 14 over 12.11.

MCGallaspy:
Fixes #95. This is only so that the user has pip.exe available, so it's not a hard requirement. I know we wish to support Python 2.6 still, but given that 2.7.10 is bundled in the installer and apparently works on all supported versions of Windows, I felt it was ok.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -322,13 +322,13 @@ procedure HandlePythonSetup;
 var
     installPythonErrorCode : Integer;
 begin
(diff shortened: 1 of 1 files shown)


aronasorman:
Well if we already bundle it, I see why not.

aronasorman:
Looks good. Merge at will.

PR #117, opened 2015-07-24
Unpack assessment items in appropriate system-wide location

MCGallaspy:
Fixes https://github.com/learningequality/ka-lite/issues/4141 for Windows.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -37,8 +37,10 @@ Name: "english"; MessagesFile: "compiler:Default.isl"
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"; Flags: unchecked
 
 [Files]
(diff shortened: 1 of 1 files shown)


benjaoming:
I don't know if this is useful, and it's just an input that you would know the relevance of to the windows installer.

If you put the assessment items in `os.path.join(kalite.ROOT_DATA_PATH, 'assessment', 'khan')`, then they will be picked up automatically as "pre-installed" assessment items, and treated and stored independently from user data. It will have the effect that running `kalite manage setup` will skip assessment items, stating that it has a pre-installed source.

The behaviour regarding the "conflict" of user-downloaded assessment items vs. pre-installed are to be determined, but keeping them in a separate place *could* be useful.

MCGallaspy:
> If you put the assessment items in os.path.join(kalite.ROOT_DATA_PATH, 'assessment', 'khan')

That's precisely what this PR is doing.

MCGallaspy:
The thing is, is that `os.path.join(kalite.ROOT_DATA_PATH, 'assessment', 'khan')` is _different_ from the as-source location... `kalite/assessment/khan` vs `kalite/content/assessment/khan`.

benjaoming:
Right cool, I was getting confused because the Windows installer still uses the single-directory layout.

PR #141, opened 2015-08-19
OS X application performance issues

### LINKED ISSUE #129: OSX application performance issues.
**Branch** master

**Target** 0.15.x

**Expected Behavior** The application should be responsive and handle calls to the `kalite` executable in a non-blocking way.

**Current Behavior** Calls to the `kalite` executable use the blocking `system` function.

Examples

1. When the `Apply` button is clicked, the app runs `kalite manage setup` which could take a long time.  This makes the user wait for about a minute or two and shows the "beach ball" mouse cursor which gives the impression that the app has "frozen".
2. We check the status of the `kalite status` to show the appropriate icon on the system menu bar.  This could manifest a "mini-freeze" while the command is being run.

**Possible Solution/s** Use the [NSTask](https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSTask_Class/index.html) class.

Since this is going to be a non-blocking call, we must somehow show the user an indicator that something is being run on the background.  Then hide that after the processing is done.

**Functionalities to check**

* [ ] Use `NSTask` class instead of the `system` function.
* [ ] Set the `KALITE_PYTHON` environment variable and point it to the `pyrun` binary for every call to `NSTask`.  This makes sure that calls like `kalite status` will work and not crash the app.
* [ ] Show an indicator (progress bar, etc) during run of command.
* [ ] Show the logs / output of the `kalite` command in the GUI or Console app.

**Related Issues**
* [KA-Lite Monitor first launch is taking a long time.](https://github.com/learningequality/installers/issues/120)

**References**
* [StackOverflow's - Execute a terminal command in a Cocoa app](http://stackoverflow.com/questions/412562/execute-a-terminal-command-from-a-cocoa-app)
* [NSTask Tutorial for Mac OS X](http://www.raywenderlich.com/36537/nstask-tutorial)

amodia:
Hi @aronasorman this will fix #129

We used `NSTask` class to run a subprocess when running kalite commands.

We are still working on logging to know that the kalite app is still processing. 

Don't merge this `PR` yet, until we finish working on kalite logging.


amodia:
Hi @aronasorman we are done working on logging.  But we need to test if this will work on `dmg` build. So don't merge this on master until we verify that it's working.

amodia:
Hello @aronasorman this is the screenshot showing on the preferences that has a log tab. This will show all the ka-lite command's output to know that the app is still processing.

screen shot 2015-09-02 at 7 32 16 pm

PR #171, opened 2015-09-18
Force restart after install

### LINKED ISSUE #123: Windows - assessment items not available immediately after installation
Using the windows installer, assessment items (and potentially other resources) may not be available immediately after installation. This is because the installer updates an environment variable (`KALITE_ROOT_DATA_PATH`), but other applications are not made aware of this change until the installer exists. In particular, it's possible to launch the app _before_ the installer exits (there's a checkbox on the last page of the wizard), in which case it will not be aware of the updated environment,

Exiting and restarting the task-tray program should make assessment items appear available, as will restarting the computer -- then all applications will be aware of the updated environment.

MCGallaspy:
Fixes #123. Turned out to be really simple. Forces a restart after installation. Does not give the option of starting the app immediately.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -29,6 +29,7 @@ SolidCompression=yes
 PrivilegesRequired=admin
 UsePreviousAppDir=yes
 ChangesEnvironment=yes
+AlwaysRestart=yes
 
 [Languages]
 Name: "english"; MessagesFile: "compiler:Default.isl"


MCGallaspy:
@cpauya can you try this out, too? I'll upload an installer and share the link here.

MCGallaspy:
See here: http://field.learningequality.org/KALiteSetup-0.15.0.exe

djallado:
@MCGallaspy I tested the the installer in windows7 and after installing, it forces the user to restart the machine.
test_exe

Before after the installation, it will automatically launch the app.
old_exe

amodia:
@MCGallaspy I also tested http://field.learningequality.org/KALiteSetup-0.15.0.exe. The same behavior with @djallado it has a default option to restart the app after the installation.
sh1

Issue #178, opened 2015-09-24
 Documentation page is not loaded in IE11 on Win8.1


arceduardvincent:
**Branch:** 0.15.x.

**Windows installer:** kalite-setup-0.15.0--9-18-2015.exe

**Platform:** IE11 on Windows 8.1

**Expected Behavior:** Fix the Documentation page returns 404 Not found. 

**Current Behavior:** 
When you navigate to the Documentation link, the documentation page returns 404 Not found. 
The path `/static/html/index.html` was not found
![demo-open-documentation](https://cloud.githubusercontent.com/assets/4099119/10068433/e5380224-62d6-11e5-96b2-0df916471a54.gif)

**Steps to reproduce:**
>1.  Navigate to Login page and login to admin.
>1.  Navigate to the user account tab then click the `Documentation` link.
>1. Notice that the Documenation page return "404 not found".

MCGallaspy:
Does a restart resolve this? Were you prompted to restart after completing installation?

MCGallaspy:
Can you also paste the traceback here?

arceduardvincent:
Hi @MCGallaspy, Yeah it prompt me to reboot.

Installer traceback error:
```python
Traceback (most recent call last):
  File "C:\Program Files\KA Lite\ka-lite\python-packages\cherrypy\_cprequest.py", line 656, in respond
    response.body = self.handler()
  File "C:\Program Files\KA Lite\ka-lite\python-packages\cherrypy\lib\encoding.py", line 188, in __call__
    self.body = self.oldhandler(*args, **kwargs)
  File "C:\Program Files\KA Lite\ka-lite\python-packages\cherrypy\_cpdispatch.py", line 34, in __call__
    return self.callable(*self.args, **self.kwargs)
  File "C:\Program Files\KA Lite\ka-lite\python-packages\cherrypy\_cptools.py", line 170, in handle_func
    raise cherrypy.NotFound()
NotFound: (404, "The path '/static/html/index.html' was not found.")
```

djallado:
I also observed the using IE10 in windows 7, and my installer is from: http://field.learningequality.org/KALiteSetup-0.15.0--9-25-15.exe.

here's my snapshot:

![screen shot 2015-09-29 at 4 36 43 pm](https://cloud.githubusercontent.com/assets/8664071/10159079/a4eb9c9e-66c8-11e5-99bc-d619b334622c.png)

It works fine using IE10 on windows 8.

[...]

PR #188, opened 2015-09-29
Show output from setup command

MCGallaspy:
Trying to debug https://github.com/learningequality/installers/issues/178.

@cpauya can you build an installer from this branch, and try it out to determine:

1. *If* the setup command is failing on ~~Windows 7 or lower~~ any version of Windows, as I suspect is happening in #178.
2. If it is failing, why?

For faster building/installing, don't bother to include assessment items, but *do* include documentation. I'm taking a crack at it myself, but given the slow build/install time it would be helpful to have many people trying it out on different windows versions simultaneously.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -416,7 +416,7 @@ begin
     Exec(ExpandConstant('{cmd}'), '/S /C "xcopy "' + ExpandConstant('{app}') + '\ka-lite\kalite\database" "%USERPROFILE%\.kalite\database\" /E /Y"', '', SW_HIDE, ewWaitUntilTerminated, retCode);
     MsgBox('Setup will now configure the database. This operation may take a few minutes. Please be patient.', mbInformation, MB_OK);
     setupCommand := 'kalite manage setup --noinput --hostname="'+ServerInformationPage.Values[0]+'" --description="'+ServerInformationPage.Values[1]+'"';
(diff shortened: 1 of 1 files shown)


MCGallaspy:
I have observed that all the static files have been copied to the user's data directory except for the docs. This leads me to believe `collectstatic` is executing without error, but that in the settings the `STATICFILES_DIRS` variable does _not_ include the location of the docs. Since `IS_SOURCE` would be `False`, following the assignments brings us to the `ROOT_DATA_PATH` setting. This is determined by an environment variable, which is set by the installer. The gotcha is that the environment isn't updated until after the installer exits, and this is what I now think is causing the docs to not be collected.

My proposed solution: remove the `setup` command responsibility from the windows installer, so that it's run the first time the server is started -- comes at the cost of making the first server start slower (bottleneck appears to be the `generatekeys` mgmt command), but with the benefit of being very easy to implement and reducing the complexity of the InnoSetup script.

@cpauya @jamalex @aronasorman agree on this?

cpauya:
Yes, let's remove the `setup` command from the installer like what we do now on the OSX app.  It's gonna take longer to run the server first time indeed but it takes one responsibility from the installer.

amodia:
Hi @MCGallaspy I tested this PR on windows 7 32 bit and I wait almost 30 minutes to start the server but it's not starting.

screen shot 2015-09-30 at 9 24 31 pm

MCGallaspy:
Okay, disregard this PR. As I mentioned in the top comment, I only wished to try to determine if the `setup` management command that is run as part of the installation process was somehow failing. Now see #189.

Issue #195, opened 2015-09-30
kalite.bat can't accept user input


benjaoming:
moved from learningequality/ka-lite#3944

We have a bunch of different kalite.bat logic :) It should probably in the end be part of the Windows installer.

Currently, one is in ka-lite/bin/ and one is generated automatically by setup.py

Original post by @MCGallaspy :

On windows 8, in develop:

Scripts which require user input (like the setup management command without the no-interaction flag) can't accept input when you call them with "kalite.bat manage mycoolscript". Works ok with "python.exe kalite mycoolscript". Since the main commands (start, stop, status) all don't require input, I'm not considering this a release blocker.

MCGallaspy:
I was referring to the file included in the ka-lite repo: https://github.com/learningequality/ka-lite/blob/0.15.x/bin/windows/kalite.bat

I don't know if this is the case with the one generated by setup.py.

benjaoming:
The one generated by setup.py is generated through an example I found on some very thorough blog post some time back, it's worth checking out.

MCGallaspy:
Now using the setup.py generated batch script.

benjaoming:
Cool, quick Q: If we factor out "python-packages" and "dist-packages" logic from setup.py (as I'm planning to do for 0.17), will it have any effect on the windows installer?

benjaoming:
@MCGallaspy

MCGallaspy:
Not sure what you mean exactly, can you elaborate? Basically, the Windows installer generates a "static" sdist using `setup.py sdist --static`, and the installs the kalite package at installation time using the resulting zipfile. So does it affect those steps?

benjaoming:
@MCGallaspy nope, that should be entirely safe! If you don't do anything outside of this regarding these to "data files" locations, then it's totally safe and removing it won't hurt.

Issue #48, opened 2015-05-22
Windows installer should use setup.py (setuptools)


benjaoming:
Hi @ruimalheiro 

Hopefully this will be helpful for you and in the future will help to reduce the responsibility of the Windows installer.

After installing Python + setuptools, which I guess is bundled with the installer, you can install the ka-lite-static zip package which can be found on pypi:

https://pypi.python.org/pypi/ka-lite-static

In the zip, which you can bundle with the installer, you find the typical `setup.py` script. It's installed by running `setup.py install` as all distributed Python packages are.

After installing it, if you can make `C:\Python27\Scripts` and `C:\Python27\` appear in the Windows path, we can have `kalite` in the command line. ( more info here: http://docs.python-guide.org/en/latest/starting/install/win/ )

MCGallaspy:
I'm debating whether we need to change the windows installer to use setuptools, given that it works ok for now. That said, if we do revisit the InnoSetup script for any reason, it's probably worthwhile to refactor it to use `setup.py` in order to consolidate the responsibility in one place and avoid code with duplicate functionality and language proliferation in the project. (I'm looking at you, PascalScript.)

aronasorman:
I think it's better to have all the installers install using the same mechanism, as that leads to less special-casing between platforms.

MCGallaspy:
> I think it's better to have all the installers install using the same mechanism, as that leads to less special-casing between platforms.

I agree, we need to figure out where this lies in our ordinal priority list.

benjaoming:
I don't know if this is useful, but this is how debian packaging works:

setup.py is invoked to install in a relative directory at build time. So setup.py is actually not run while installing the .deb, but while building it!

This is perfectly fine, as the .deb format is intended to contain files to be placed on the target system which are then managed by dpkg.

Would a similar install builder for Windows exist? Otherwise, would it be possible to put all of the Windows installation logic in a separate setup_windows.py that invokes/imports setup.py and thus contain everything for Windows in the original source tree without a separate software project? Recall that a setup.py can be made into a setup.exe with distutils...

MCGallaspy:
Could be useful, seems like we could do essentially the same thing on Windows -- it's just a matter of rewriting the scripts and debugging them. We also manage some upgrade scenarios through InnoSetup, so we'd have to figure out how we wish to handle that.

PR #218, opened 2015-10-13
Use python setuptools to unpack kalite source

MCGallaspy:
A WIP. Will ultimately fix https://github.com/learningequality/installers/issues/195 and finally also https://github.com/learningequality/installers/issues/48.

First commit message:
```text
Previously, we created an archive of the kalite source (and its dependencies)
but because I didn't know anything about setuptools I wasn't using it properly.

Now instead of copying the whole source, we reduce the responsibility of
the windows installer by copying the source archive and unpacking it using
git. This means that the source now lives in the python site-packages instead
of the user-chosen installation directory. That directory now just has the
source archive and systray executable files.

We no longer set KALITE_ROOT_DATA_PATH, since it will now be unpacked
in the expected location by pip (namely sys.prefix + "share\kalite").

Since we don't set any environment variables, don't force a restart either.

The systray executable will need to be made aware of the location of the
`kalite.bat` script, since it no longer lives in the application directory.

We may discover more tasks as development continues.
```


benjaoming:
@MCGallaspy would this be a viable success criteria?

*After I install kalite using the Windows installer, I can update the kalite core using `pip install ka-lite-static --upgrade`*

Not saying it's a realistic scenario, but just that it's an interesting synergy from a technical perspective.

MCGallaspy:
Yes, that's totally realistic. I've tried to outline my intent above, but yeah basically the windows installer is just installing ka-lite-static to the user's Python site-packages using pip.

MCGallaspy:
As well as handling upgrade scenarios and providing a python installer.

benjaoming:
By realistic scenario, I meant that someone installing with the Windows installer isn't very likely to open up a command line and do a pip install afterwards :)

MCGallaspy:
Installer now needs to set KALITE_SCRIPT_DIR environment variable.

MCGallaspy:
This is ready to merge. Will upload an installer to test. Note to testers: This just touches the installation process. The executable I'm uploading makes no guarantee about the usability of the KA Lite app (it's a dev version) and does not include assessment items. The thing to test is whether it installs correctly and the server can start/stop.

MCGallaspy:
@benjaoming this went much more smoothly than I anticipated, btw. Good job on the setuptools stuff!

benjaoming:
That's good news! Let's hope that we can keep up putting all the value in the common "sdist" pot! Optimistic about the future... Windows may not be that bad :)

windows/gui-source/KA Lite/KALite.cpp
```diff
@@ -16,17 +16,45 @@ fle_TrayMenuItem * menu8;
...
+		window->sendTrayMessage("KA Lite", "Error: Environment variable KALITE_SCRIPT_DIR is not set.");
+		buffer = 0;
+	} 
+	else if (bufsize > MAX_SIZE)
+	{
+		char err_message[255];
+		sprintf(err_message, "Error: the value of KALITE_SCRIPT_DIR must be less than %d, but it was length %d. Please start KA Lite from the command line.", MAX_SIZE, bufsize);
+		window->sendTrayMessage("KA Lite", err_message);
+		buffer = 0;
+	}
```
benjaoming: some rock solid stuff there!

windows/gui-source/KA Lite/KALite.cpp
```diff
@@ -16,17 +16,45 @@ fle_TrayMenuItem * menu8;
...
 void startServerAction()
 {
-	if(!runShellScript("kalite.bat", "start", "ka-lite\\bin\\windows\\"))
+	const DWORD MAX_SIZE = 255;
+	char script_dir[MAX_SIZE];
+	kaliteScriptPath(script_dir, MAX_SIZE);
+	if(!runShellScript("kalite.bat", "start", script_dir))
 	{
 		// Handle error.
-		printConsole("Failed to run the script.\n");
```
benjaoming: Not sure about the consequence of this? It looks like if the start command fails, it does so silently? Is that because there's a log file, anyways and then it's all fine?

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -38,10 +37,8 @@ Name: "english"; MessagesFile: "compiler:Default.isl"
 Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{cm:AdditionalIcons}"; Flags: unchecked
 
 [Files]
-Source: "..\ka-lite\*"; DestDir: "{app}\ka-lite"; Excludes: ".KALITE_SOURCE_DIR,content\assessment,content\assessment\*,content\assessmentitems.sqlite"; Flags: ignoreversion recursesubdirs createallsubdirs
-Source: "..\ka-lite\content\*"; DestDir: "{app}\ka-lite\content"; Excludes: "assessment,assessment\*,assessmentitems.sqlite"; Flags: ignoreversion recursesubdirs createallsubdirs
-Source: "..\ka-lite\content\assessment\*"; DestDir: "{app}\ka-lite\assessment"; Excludes: "assessmentitems.sqlite"; Flags: ignoreversion recursesubdirs createallsubdirs
-Source: "..\ka-lite\content\assessmentitems.sqlite"; DestDir: "{app}\ka-lite\assessment\khan"; Flags: ignoreversion
+Source: "..\ka-lite\dist\ka-lite-static-*.zip"; DestDir: "{app}\ka-lite"
```
benjaoming: I'm always releasing both a tar.gz and a .zip, the latter I wasn't sure about if was required, but I take it that from now on, we should continue that :)

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -53,15 +50,12 @@ Name: "{group}\{cm:ProgramOnTheWeb,{#MyAppName}}"; Filename: "{#MyAppURL}"
 Name: "{group}\{cm:UninstallProgram,{#MyAppName}}"; Filename: "{uninstallexe}"
 Name: "{commondesktop}\{#MyAppName}"; Filename: "{app}\{#MyAppExeName}"; Tasks: desktopicon ; IconFilename: "{app}\images\logo48.ico"
 
-[Run]
-Filename: "{app}\{#MyAppExeName}"; Description: "{cm:LaunchProgram,{#StringChange(MyAppName, '&', '&&')}}"; Flags: shellexec postinstall skipifsilent
-
 [Dirs]
-Name: "{app}\"; Permissions: everyone-modify
+Name: "{app}\"; Permissions: everyone-readexec
```
benjaoming: Keep an eye on the assessmentitems.sqlite database, which I had to do a quickfix for in the .deb -- because I put assessment items in a system wide location, where django then coughed because the sqlite db wasn't readable. This should be fixed in the sdist, anyways.

https://github.com/learningequality/ka-lite/issues/4638

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -528,15 +521,14 @@ end;
 
 { Called just prior to uninstall finishing. }
 { Clean up things we did during uninstall: }
-{ * Remove environment variable KALITE_ROOT_DATA_PATH }
+{ * Remove environment variable KALITE_SCRIPT_DIR, which is set starting in version 0.16.x }
+{ * Previously (versions 0.13.x to 0.15.x) KALITE_ROOT_DATA_PATH was set -- it should be unset by the respective }
+{   uninstallers of those versions }
```
benjaoming: You could potentially also unset it in this installer, just to be sure? :)

benjaoming:
I tried firing some questions off to make it look like a review, but I'm no expert here! Looks like really good work from my POV, and the questions are just food for thought... but perhaps a little undercooked so no need to respond if you think it's banal :)

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -29,7 +29,6 @@ SolidCompression=yes
 PrivilegesRequired=admin
 UsePreviousAppDir=yes
 ChangesEnvironment=yes
-AlwaysRestart=yes
```
aronasorman: We no longer need to restart the app upon install?

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -342,15 +336,24 @@ begin
...
+    if not ShellExec('open', PipPath, PipCommand, '', SW_SHOW, ewWaitUntilTerminated, ErrorCode) then
     begin
       MsgBox('Critical error.' #13#13 'Dependencies have failed to install. Error Number: ' + IntToStr(ErrorCode), mbInformation, MB_OK);
       forceCancel := True;
       WizardForm.Close;
     end;
+
+    { Must set this environment variable so the systray executable knows where to find the installed kalite.bat script}
+    { Should by in the same directory as pip.exe, e.g. 'C:\Python27\Scripts' }
+    RegWriteStringValue(
```
aronasorman: Do these registry values get removed upon uninstall?

aronasorman:
Since I'm not an expert in InnoSetup/Windows/C++ myself, I don't have much else to test here. Is there an installer that @jtamiace or @EdDixon can test?

windows/gui-source/KA Lite/KALite.cpp
MCGallaspy: Seems to have been a debugging line that was left in -- the user won't see this output normally.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: `python setup.py sdist --static` on Windows defaults to a zipfile.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: Okay, I'll test out if this is an issue.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: Sure.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: No, turns out we can just disable the option to start the program directly from the installer.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: Yeah.

MCGallaspy:
> Is there an installer that @jtamiace or @EdDixon can test?

Yeah, it's linked above: https://github.com/learningequality/installers/pull/218#issuecomment-148850789

windows/gui-source/KA Lite/KALite.cpp
MCGallaspy: But yeah, there's no catch-all failure notification right now. Wasn't sure what kind of feedback would be appropriate. Note that the most likely failure is probably from the `kaliteScriptPath` function which already sends a notification to the systray executable, and subsequent notifications will trample prior ones before the user can read them.

EdDixon:
In testing I discovered that at server start 2 python threads do start but immediately shut down and the server itself never starts.

Immediately after server start:




A few moments later:

MCGallaspy:
What version of windows was that @EdDixon ?

EdDixon:
Oh sorry XP

MCGallaspy:
@EdDixon I think failure to start may be my error, due to a missed step in the installer build process. However, I think the fix here is good in that in successfully starts the process for the server -- the fact that the server then chokes is a separate issue.

To confirm that this is the case, I suggest.

1. Install KA lite using the above installer. Don't start the server.
2. Type the following commands after installation:
    1. `cd %KALITE_SCRIPT_DIR%` -- this will take you to where `kalite.bat` lives
    2. `kalite.bat manage init_content_items --overwrite`
3. Then start the systray program and try starting the server. If I'm correct, it should now work. Methinks the missing ingredient from the build process is step 2.

MCGallaspy:
I tried the above sequence -- didn't quite work, but at any rate it's definitely an application bug. The title fix is good.

benjaoming:
@MCGallaspy this could be part of "make sdist" (and the step is already
included) -- could the build process use the Makefile? Methinks if you need
changes to make it (cyg)win-friendly, that's just fine.. there would be
some benefit in a unified build command.

MCGallaspy:
@benjaoming yeah, I could use the Makefile -- but namely the above fix has to do with the installer, and works fine. Whatever mistake I made in the build process here obscures the main point of the PR.

radinamatic:
@MCGallaspy  - Results of my tests with the above installer:

**XP**

Terminal output during the installation (took quite a while):
pip

Once the installation finished, it did not ask for restart, so I started with the desktop shortcut. Systray executable appears correctly, but does not manage to start the server (note no `python.exe` in the Task Manager view):

no-python

`echo %KALITE_SCRIPT_DIR%` gives `G:\Python27\Scripts` which indeed contains `kalite.bat` file.

Running `kalite.bat status` gives the following output:
```
G:\Python27\Scripts>kalite.bat status
G:\Python27\python.exe: can't open file 'G:\Python27\share\kalite\kalitectl.py': [Errno 2] No such file or directory
```
Indeed, no `kalitectl.py` file is to be found in `G:\Python27\share\kalite\kalitectl.py` folder.

I had various KA Lite versions running on this XP machine up to the latest 0.15 release installed last week.  



**Windows 8**, OTOH, all running smoothly, screenshots in a moment...

radinamatic:
**Windows 8**, terminal output (installation took less time than on XP):

win8b


Once the installation finished, I started KA Lite with the desktop shortcut, and the server started from systray almost immediately and VERY fast. Note 2 python.exes in the Task Manager view.

win8a

radinamatic:
Will test in Windows 7 later if necessary...

MCGallaspy:
@radinamatic no restart expected now. Don't necessarily expect the server to *start*, since this PR targets the installer and systray executable, and the included KA Lite is a dev version. As long as the systray doesn't notify of an error and a python process is *started* (even if it subsequently dies before the server starts) then we'll call that a success here. The concerning part on the XP install you mentioned is that `kalitectl.py` can't be found. Will take a look.

Would be good to give the same check of Windows 7 that you did here.

radinamatic:
> @radinamatic no restart expected now. 

Gathered as much from the missing requirement on last installer window.

> Don't necessarily expect the server to start, since this PR targets the installer and systray executable, and the included KA Lite is a dev version. As long as the systray doesn't notify of an error and a python process is started (even if it subsequently dies before the server starts) then we'll call that a success here. 

Systray itself did not notify any errors, just got stuck without managing to start the server, probably because of:

> The concerning part on the XP install you mentioned is that kalitectl.py can't be found. Will take a look.

> Would be good to give the same check of Windows 7 that you did here.

Will to as soon as I finish something else and reboot.

radinamatic:
@MCGallaspy, Windows 7 test went smoothly, almost as on W8: `pip.exe` terminal output was literally the same, installation finished with no issues, although I wasn't able to start the server from systray at first (`python.exe` process appeared briefly, than disappeared). However, once I restarted the machine itself, I managed to start the KA Lite server too with no issues.

MCGallaspy:
@radinamatic can't replicate your error on XP. Can you describe any other actions you may have taken prior to installing (besides seemingly installing 0.15.0)? Had you been running a development version from that box?

MCGallaspy:
I recommend merge, then see if we can turn up @radinamatic's issue in later testing.

radinamatic:
@MCGallaspy Yes, I had the development version on this box until the `browserify` refactor (mid July), and since then I probably installed and tested all the pre-release installers for 0.14 and maybe 5 for 0.15.

MCGallaspy:
Rebased onto master

MCGallaspy:
Closed to rebase onto develop. See #262.

PR #222, opened 2015-10-13
0.15 debian changes

### LINKED ISSUE #146: Software Center issues
Several issues pertaining Software Center installations...

 * [x] @radinamatic doesn't see the below screen at all because of an Ubuntu bug
 * [x] Add documentation about Software Center bug because it can lockup the UI
 * [x] The `$SUDO_USER` doesn't seem to be reflected when running debconf through Software Center.
 * [x] Username was only prompted when selecting the system service init.d script

screenshot from 2015-08-27 00 47 58

### LINKED ISSUE #191: Alternatives for writing downloaded assessment items to /tmp
Writing assessment.zip to /tmp or other location can cause problems for small devices.

We need this to be configurable with debconf, e.g.:

```
# Set debconf before installing the package
echo "ka-lite-raspberry-pi ka-lite/download-tmp-folder select /other/path" | debconf-set-selections
# (...)
```

Furthermore, we should measure disk space and raise a proper warning or failure if it isn't sufficient, even though wget will also do that. The difference should be that our failure message tells the user who to configure an alternative location for the download or:

> maybe you could check the disk space, and if isn't availabe prints how to download and install it later. It would make the installation smooth.

### LINKED ISSUE #221: An empty username makes the installer fail
Fix coming in...

### LINKED ISSUE #223: Add manpage for kalite command
:)

@cpauya I accidentally did not include kalite.1.gz in the sdist because I forgot to list it in MANIFEST.in -- so in the Debian, I'll add it manually. There'll be a patch update to fix it.

benjaoming:
WIP

Incoming changes for 0.15 debian tracked here.

Not sure what the develop and master branch are for.

Fixes #221 
Fixes #191 
Partly fixes #146
Fixes #223


benjaoming:
For anyone wondering, this PR will be closed and re-submitted as a single commit :)

MCGallaspy:
Can you re-target develop after I merge #254?

benjaoming:
Yes, np.. not sure why this would be in a separate develop branch, but there are no problems.

It's not going in until released in the PPA. It's just tracking changes until released in some beta or 0.15-ubuntu2 release.

MCGallaspy:
It's just to consolidate our development efforts into one branch. See here: https://github.com/learningequality/installers/pull/250#issuecomment-153552260

benjaoming:
Not very happy about loosing the #222 number :(

PR #234, opened 2015-10-20
Copy 0.13.x files directly to .kalite dir

### LINKED ISSUE #179: Upgrading using windows installer can take a very long time
Encountered this when installing 0.15 on top of 0.13.2. If an old installation is detected, it will attempt to migrate data files from old location in the source directory to the .kalite directory -- but this takes a _really_ long time. I suspect it's due to the huge number of assessment item files. I have two ideas:

* Give some feedback during the process, by showing the cmd.exe window handling this process instead of suppressing it. This creates the possibility that a user might prematurely terminate that process, though, which could result in unrecoverable data loss.
* It's probably unnecessary to copy assessment item files, so the process could be tweaked to be more selective.

MCGallaspy:
Edit: A WIP. Need to test this out comprehensively.

We avoid copying the files to a temp location first, since it could require
a lot of storage space and take a long time.

Fixes #179.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -83,15 +83,12 @@ var
   restoreDatabaseTemp : integer;
   forceCancel : boolean;
   prevVerStr : string;
-  prevAppBackupDir : string;
-  runGitmigrate : boolean;
 
 procedure InitializeWizard;
 begin
     isUpgrade := False;
     forceCancel := False;
-    runGitmigrate := False;
-    
+
     if WizardForm.PrevAppDir <> nil then
     begin
         ShellExec('open', 'taskkill.exe', '/F /T /im "KA Lite.exe"', '', SW_HIDE, ewWaitUntilTerminated, stopServerCode);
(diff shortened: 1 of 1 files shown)


MCGallaspy:
Tested on Windows 7. Feel confident about it, but I need to test it out on XP. It's come to my attention that xcopy may not be available on XP?

MCGallaspy:
`xcopy` is available on Windows XP. IMO the fix is good. Encountered the same issue as https://github.com/learningequality/installers/pull/218#issuecomment-149738405, but that is seemingly due to a missing build step.

MCGallaspy:
@aronasorman or @benjaoming or @cpauya, can someone review this? Some notes on my changes to make evaluation easier will follow, we can also schedule a chat.

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -196,42 +193,20 @@ begin
     end;
 end;
 
-procedure Backup013Files;
```
MCGallaspy: Delete this function. We no longer wish to copy files to a temporary location first -- it could potentially waste a lot of disk space. Instead, we just directly call the `DoGitMigrate` function below (which could have a better name?) in order to copy video files and the database from the previous app directory directly to the user's `.kalite` directory.

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -196,42 +193,20 @@ begin
...
-    begin
-        MsgBox('Fatal error' #13#13 'Failed to backup database directory.', mbError, MB_OK);
-        forceCancel := True;
-        WizardForm.Close;
-    end;
-    Exec(ExpandConstant('{cmd}'), '/S /C "xcopy "' + prevContentDir + '" "' + prevAppBackupDir + '\content\" /E"', '', SW_HIDE, ewWaitUntilTerminated, resCode);
-    if resCode <> 0 then
+    MsgBox('Migrating old data to current user''s %USERPROFILE%\.kalite\ directory.', mbInformation, MB_OK);
+    Exec(ExpandConstant('{cmd}'), '/S /C "mkdir "%USERPROFILE%\.kalite""', '', SW_SHOW, ewWaitUntilTerminated, retCode);
+    Exec(ExpandConstant('{cmd}'), '/S /C "xcopy "' + WizardForm.PrevAppDir + '\ka-lite\content" "%USERPROFILE%\.kalite\content\" /Y"', '', SW_SHOW, ewWaitUntilTerminated, retCodeContent);
```
MCGallaspy: These commands have been modified slightly -- instead of recursively copying the entire source directory (which would include assessment items) we just do a shallow copy of the old content directory. All the videos live (flat) in that directory, which is what we wish to copy.

MCGallaspy: I also changed them to `SW_SHOW`, which will show the actual command prompt as the process is executed. This gives some feedback to users. Prior to this, if the copy was long running there was no indication -- the wizard just sits there with the progress bar full, but unable to be used.

cpauya:
Hi @MCGallaspy - Sorry took a long while to respond here.  I'm still setting-up my Win 7 VM (my old one already expired).  Hopefully will be done by tomorrow to be able to run and check this out. :smile_cat:

MCGallaspy:
:+1:

djallado:
@MCGallaspy I tested your PR in IE11-Win7, I compare it with 0.15.0 from https://learningequality.org/r/windows-installer-0-15. I observed that the time during upgrade process of 0.3.2 to 0.15.0 from https://learningequality.org/r/windows-installer-0-15 was the same, as I upgrade 0.3.2 to the build from your PR. Maybe I need to have more data so that I can compare the time involve during upgrade. I will test it again with more data today.

MCGallaspy:
@djallado if you want to test this out without downloading a bunch of videos, you could just put several large files in the content directory of a 0.13 install.

djallado:
I tried 3.95 GB of data in content folder, upgrading from 0.13.2 to the build from your PR is faster compare to upgrading from 0.13.2 to 0.15.0 from https://learningequality.org/r/windows-installer-0-15.

MCGallaspy:
Okay, if there are any other issues hopefully we'll turn them up during more comprehensive testing.

Issue #143, opened 2015-08-22
Remove the "monitor" in the name of the OSX installer and app.


cpauya:
**Branch** master

**Target** 0.15.x

**Expected Behavior** The name of the installer and app must only be "KA-Lite.dmg" and "KA-Lite.app" respectively.

**Current Behavior** The name of the installer and app is "KA-Lite-Monitor.dmg" and "KA-Lite-Monitor.app" 

**Solution/s** Remove the "monitor" from the name.  

Notes:

1. There may be hard-coded strings referring to "monitor" which must be double-checked in the app itself.
2. Check that the `setup.sh` script is also clean of the "monitor" word.

cpauya:
Fixed in #243.

PR #243, opened 2015-10-28
Hotfix/143 remove the Monitor in the name of the OS X installer and app

djallado:
@cpauya, this fixes the issue #143.

I removed the "Monitor" in the name of the OS X installer and app.


osx/KA-Lite/KA-Lite.xcodeproj/project.pbxproj
```diff
@@ -79,13 +79,13 @@
 		F1DDD40D1A6E5DC10022F1C9 /* Products */ = {
 			isa = PBXGroup;
 			children = (
-				F1DDD40C1A6E5DC10022F1C9 /* KA-Lite-Monitor.app */,
-				F1DDD41F1A6E5DC10022F1C9 /* KA-Lite-MonitorTests.xctest */,
+				F1DDD40C1A6E5DC10022F1C9 /* KA-Lite.app */,
+				F1DDD41F1A6E5DC10022F1C9 /* KA-LiteTests.xctest */,
```
cpauya: We used `KA-Lite-Tests` above but you are using `KA-LiteTests` here @djallado - we should be consistent.

cpauya:
This is very good!  I will merge this in so we can resume with all outstanding issues.  Thanks for this @djallado! :smile_cat:

osx/KA-Lite/KA-Lite.xcodeproj/project.pbxproj
cpauya: @djallado Let's open an issue about this if it presents an issue.

PR #250, opened 2015-11-03
Develop into master

MCGallaspy:
Hi @amodia and @cpauya, et al. In order to consolidate our efforts, can we consider merging develop into master and continuing development on master? Then, when we release a new version, we can tag it and/or create a new branch to support it. @aronasorman thoughts?

I'm open to other suggestions, e.g. doing something similar to ka-lite, where we work on a develop branch and master always reflects the latest release or something. But simpler is better IMO, as long as we have a consensus.


cpauya:
Hi @MCGallaspy - we've been using the `develop` branch mainly for the following reasons:

1. Consistency - since the `ka-lite` repo is using the same approach.
2. [A successful Git branching model by Nvie](http://nvie.com/posts/a-successful-git-branching-model/) -- on principle that the `master` branch must be deployable upon checkout and that the `develop` branch must contain all features/hot fixes under development.

Why don't we use tags to "tag" our releases instead of branches as per the Nvie model above?

MCGallaspy:
Very well, I'll merge master into develop then, and re-target my existing PRs to develop.

cpauya:
Thanks much! :+1:

benjaoming:
Master isn't actually deployed but I can still go with this :) Maybe we will deploy master some day where everything is crazy automated and Bamboo is building everything upon just the slightest indentation change.

:+1:

PR #260, opened 2015-11-06
Merged develop into master

amodia:
@cpauya I merged develop into the master branch.


djallado:
@amodia, I clone from installer repo and checkout your branch so that I can test it in clean source. When I open it on `Xcode` there's no error because pyrun-2.7 directory is already in the Resources` directory. 
When I run `setup.sh` it builds `KA-Lite.pkg` in the `output` directory without error.

cpauya:
Tested thus merging.

PR #262, opened 2015-11-06
Use python setuptools for windows installer

### LINKED ISSUE #48: Windows installer should use setup.py (setuptools)

MCGallaspy:
#218 rebased onto develop. Fixes #48 and #195.


MCGallaspy:
I'm tempted to make a unilateral decision about this PR.

cpauya:
Sorry @MCGallaspy - will check this out today using our VMs.

cpauya:
This is the same as #218 targeting `develop` branch, so merging.

PR #285, opened 2015-11-30
modify nginx requirements

benjaoming:


benjaoming:
Was missing a "git push" just before merging #263 

This was also released as 0.15.0-ubuntu4

PR #287, opened 2015-12-01
Windows -- Unpack assessment items during install

MCGallaspy:
Submitted after mucking around with InnoSetup for some time.

Commit message:

```text
Instead of unpacking assessment items when creating the installer, which
results in a metric boat load of files (and causes wine+IS to choke) we
instead include khan_assessment.zip in the installer, and call
the unpack_assessment_zip mgmt command during the installation process.

We *must* reference the KALITE_SCRIPT_DIR environment variable directly
from the registry -- the environment for any process spawned by the
installer won't be updated until *after* the installer exits, so e.g.
%KALITE_SCRIPT_DIR% won't expand as expected.
```


MCGallaspy:
Merging.

PR #300, opened 2015-12-03
Don't copy empty db; ka-lite app now handles it

### LINKED ISSUE #297: Windows upgrade from 0.15.0 to 0.15.1 removes account data
While testing the 0.15.1 installer and upgrade process from 0.15.0 I encountered on metal and replicated in VM that all accounts are lost including the admin account during an upgrade from a previous 0.15.0 install. This occurred on Windows 10 and has only been replicated in that environment so far. I have not looked at the contents of the remaining SQLite file to see if just the accounts are removed or all data in the database file, but it does appear from the available UI that all data is has been removed.

1)To replicate Install KA-Lite 0.15.0
2)Create Admin, coach, and user accounts
3) do some exercises to populate the student account
4) run the current 0.15.1 installer to do the upgrade
5) during that prior step there is a dialog showing the uninstallation of the previous 0.15.0 install at the very next dialog with options for the startup configuration you can navigate to the .kalite/database folder and verify that the database file is still there 
6) proceed with completion of the upgrade
7) you will discover the admin account has to be created again so go ahead and do that (with exactly the same credentials if you want)
8) you will also find the device needs to be registered again go ahead and do that using the same method as previous as well if desired(in my case I was using one click registration methods)
9) Go to the manage default facility tab and you will see the accounts you created are no longer there either. 

It may be that the student data does still exist in the database file and the relational association's are now broke, at this point I do not know. I am trying to examine the database itself now to see if those records still exist.

MCGallaspy:
Fixes #297 on develop. The ka-lite app now handles copying a pre-built database (if it exists).

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -368,14 +368,12 @@ begin
   result := True;
 end;
 
-{ Completes the setup with bundled empty database file. }
 procedure DoSetup;
 var
     retCode: integer;
 begin
-    { Copy the bundled empty db to the proper location. }
     { Used to have more responsibility, but we delegated those to the app itself! }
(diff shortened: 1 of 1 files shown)


MCGallaspy:
See https://github.com/learningequality/ka-lite/pull/4656 for how it's now handled within the app.

PR #319, opened 2015-12-16
Bamboo make script for Windows installer

MCGallaspy:
Goes hand in hand with https://github.com/learningequality/ka-lite/pull/4762. Used to be an inline script on the build server, but I was reduced to trying different things one-by-one to see what breaks the build and what doesn't. Doesn't seem to have made a difference... ?

Don't merge, opened to keep track.

### DIFF
--- a/windows/make_dist.sh
+++ b/windows/make_dist.sh
@@ -1,0 +1,12 @@
+pushd ka-lite
+
+/usr/local/bin/virtualenv venv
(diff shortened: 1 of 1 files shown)


MCGallaspy:
If this build succeeds, then this patch doesn't make a difference: http://dungeon.learningequality.org:8085/browse/KL-KALITE016-80

Issue #321, opened 2015-12-17
Failed to create config file/option alerts on Windows installer


radinamatic:
"Failed to create config file/option" alerts

**Branch**: 0.16 installer (build # 81) on Windows 7 & 8 

**Current Behavior**:  

Two error alerts appear when KA Lite is started from the desktop shortcut:
![config](https://cloud.githubusercontent.com/assets/1457929/11842145/0f1ff0a8-a402-11e5-9eae-e89cbb462a56.png)
![option](https://cloud.githubusercontent.com/assets/1457929/11842151/1866e8ce-a402-11e5-80b9-ad06907719de.png)

Both errors alerts appear twice (1st, 2nd, 1st, 2nd), but after that KA Lite appears in the Taskbar notification area, and can start the server correctly.

MCGallaspy:
@ruimalheiro perhaps you can give some insight? Not really sure what the [config.h](https://github.com/learningequality/installers/blob/develop/windows/gui-source/KA%20Lite/config.h) stuff is for.

ruimalheiro:
Hey @radinamatic @MCGallaspy ,

The options to run at windows startup / autostart server  are stored within the file "CONFIG.dat". The "config.h" contains all the logic to create and read the configuration file.

Perhaps that error is related with writing permissions issues by the user? If it fails to create the file, it will display that error.

radinamatic:
So far these errors appears on Windows 7 & 8, but not on XP. Can that maybe shed some light?

ruimalheiro:
That would make sense since Windows Vista+ are more restrictive in respect to user permissions.

That would make sense since that Windows Vista+ are more restrictive in respect to user permissions.

If the user does not have permissions to write to the directory where KA Lite is installed, it will not be able to create the configuration file and it will display that error. 

Unfortunately we are not printing the errors to a log file or so. That would help a lot because we could easily check the error code under MSDN, that would provide us more detailed information. 

Could you try to run KA Lite as admin (right click on exe, run as admin) and check if the same error is happening?

radinamatic:
The user that is installing KA Lite and running has admin status, and those errors do not appear while running the `KA Lite.exe` (weather by double click, right click >`Open` or right click > `Run as administrator`). 

However, if I want to activate `Options > Run KA Lite when the user logs in`, by right clicking KA Lite icon in the taskbar notification area, I get "Failed to create config file".

MCGallaspy:
[...]

Issue #296, opened 2015-12-02
Windows Installer python version check foiled if Python 3 is also installed


EdDixon:
On Windows 10 and without the python 2.7.10 path set in the windows but with python 2.7.10 installed, an installation of 0.15.1 will error with the following error messages should the user choose to answer no knowing python is installed just not in path. The default option for the 2.7.10 python installer is not to add itself to the path string but it is optional to include that. Additionally KA-Lite can be installed without that if the user sets a specific environent variable for kalite to the required path (KALITE_PYTHON). We should not force the user to install a redundant copy of python on the system potentially causing confusion as to which python path to follow for KA-Lite and other applications. Screenshots of the related message dialogs attached. 
![2015-12-02_13-17-39](https://cloud.githubusercontent.com/assets/4349806/11543969/ade68b40-98fc-11e5-8222-fce5f21e7f52.png)
![2015-12-02_13-30-59](https://cloud.githubusercontent.com/assets/4349806/11543973/b41624da-98fc-11e5-9444-81df3762c870.png)
![2015-12-02_13-31-19](https://cloud.githubusercontent.com/assets/4349806/11543976/b8db4054-98fc-11e5-820b-baca3396d958.png)

MCGallaspy:
Did you verify as you claim that having python in the path does not cause
this issue?

EdDixon:
No I have not as that would take time and as I understand it this is for eminent release without much testing planned, however another potential bug has now appeared. During replication of it I will add the python path to the path string here during this test going on now to verify that both replicate. (I hope not)

EdDixon:
@MCGallaspy verifying now that with python in the environment path this is in deed not an issue as I have claimed.

MCGallaspy:
Do you also have python 3 installed?

EdDixon:
Python 3 is installed on this live development machine along with 2.7.10 from the previous KA Lite installation of version 0.15.0 which I was upgrading and is apparently a factor leading to the error. In a clean WIndows 10 VM upgrade test and with no other versions of Python installed, I am not prompted to rerun the python installer while only python 2.7.10 is installed from the previous 15.0 installation. Also on the development machine there is no specified python version added to the path but pathing for KA-Lite is handled by KALITE_PYTHON env var which may also have been a factor.

MCGallaspy:
>  In a clean WIndows 10 VM upgrade test and with no other versions of Python installed, I am not prompted to rerun the python installer

That's also my observation.

[...]

Issue #94, opened 2015-07-13
ka-lite-static is not installed when I install latest build from dungeon on Windows Installer.


djallado:
**Branch:** `Master (Windows Installer Build 183)`

**Expected Behavior:** `ka-lite-static` should be installed when I install the latest build from dungeo. Like in the screenshot:
![fornow](https://cloud.githubusercontent.com/assets/8664071/8649247/eedceb70-299a-11e5-927b-ddf49db56717.png)

**Current Behavior:** `ka-lite-static` is not installed after I install the latest build. 
Here's the screenshot if I pip freeze:
![latest](https://cloud.githubusercontent.com/assets/8664071/8649265/37d8cd44-299b-11e5-909d-36f5da9ac89b.png)

**Steps to reproduce:**
1. Install `IE810- Windows7` in `virtual box`
2. Download windows installer **build 183** from `dungeon`
3. Install `KALiteSetup-0.14.0.exe` *(latest installer downloaded from dungeon)*
4. Follow `KA Lite Setup Wizard`
5. Launch `KA Lite`
6. Select `Computer` from the `Start menu`
7. Choose `System Properties` from the context menu
8. Click `Advanced system settings` > Advanced tab
9. Click on `Environment Variables`, under `System Variables`, find `PATH`, and click on it.
10. In the `Edit windows`, modify `PATH` by adding `;C:\Python27;C:\Python27\Scripts`. and click ok.
11. Reopen `Command prompt` window, and type `pip freeze`. There's no `ka-lite-static` like in the screenshot:
![latest](https://cloud.githubusercontent.com/assets/8664071/8649467/2e1a0fc8-299d-11e5-8fbf-d098be9a88bd.png)

To make it work, I have to install `kalite-static` manually like screenshot below:
![after_install_static](https://cloud.githubusercontent.com/assets/8664071/8647905/b4932580-298e-11e5-80f0-fdac0acf00e6.png)

Then when I launch KA-Lite, it works fine. I can now download videos like in a screenshot below:
![123](https://cloud.githubusercontent.com/assets/8664071/8648239/976ce6fa-2991-11e5-9ba8-f43295cb0dd5.png)

aronasorman:
Known bug, and something I want to fix. Has something to do with wine, as the build in windows itself works. Somehow running InnoSetup through wine only bundles a small subset of the assessment items.

MCGallaspy:
[...]

Issue #91, opened 2015-07-10
There's a bug that always shows the user preferences dialog.


djallado:
**Branch:** `Master` (OS X Installer)
 The `User preferences` dialog always shows because database location has changed.

**Expected Behavior:**
`User preferences` dialog will disappear after clicking the `Apply` button.

**Current Behavior:**
The `User preferences` dialog always show after clicking the `Apply` button.

**Steps to reproduce:**

1. Load `KA-Lite Monotor` app.
2. Set `username` and `password` in `User preferences` dialog and click `Apply`.
![screen shot 2015-07-10 at 12 03 28 pm](https://cloud.githubusercontent.com/assets/8664071/8612000/2a06aaca-26fc-11e5-9dc4-373c419004a5.png)
3. Notice that the `User preferences` dialog is alway shown until the `Start KA Lite` menu is clickable.

**Screenshots below:**
![screen shot 2015-07-10 at 1 13 41 pm](https://cloud.githubusercontent.com/assets/8664071/8612611/d234016c-2705-11e5-89a9-c69d6953a3c5.png)
![screen shot 2015-07-10 at 1 14 57 pm](https://cloud.githubusercontent.com/assets/8664071/8612615/df61a10a-2705-11e5-96c3-7a7807f2f0c1.png)
![screen shot 2015-07-10 at 1 23 23 pm](https://cloud.githubusercontent.com/assets/8664071/8612690/3fb011d0-2707-11e5-846a-e794f2cde12f.png)

cpauya:
This is because the location of the database has changed, so when the .app is loaded and the database is not found, it will prompt for the user preferences dialog.

@djallado Can you add this to your ToDos?

PR #322, opened 2015-12-18
Add systemstart task scripts to installer repo, fix bug

### LINKED ISSUE #106: KA-Lite server does not start on boot after option selected post installation
OS : XP 
Browser : Chrome 
Version : 0.14 Development Installer

After installation during which the option to start on boot was not selected, enabling the option after installation does not have the expected affect of having the server start after a reboot of the system. The screen shot after rebooting shows the browser on 120.0.0.1:8008 after waiting and refreshing for several minutes, also no python processes have started.

MCGallaspy:
Fixes #106 -- should start at boot now.
These scripts are updated, and should be removed from the ka-lite repo... they do no good there!

commitmsg
```
Copy them, and instead of referencing the now defunct bin/windows/kalite.bat
script, use the included kalite.bat script in the ..\Python27\Scripts\ dir.
```

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -444,7 +444,7 @@ begin
       
             if StartupPage.SelectedValueIndex = 0 then
             begin
-                if ShellExec('open','guitools.vbs','4', ExpandConstant('{app}'), SW_HIDE, ewWaitUntilTerminated, StartupCode) then
+                if Exec(ExpandConstant('{cmd}'), '/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"', '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
                 begin
                     if Not SaveStringToFile(ExpandConstant('{app}')+'\CONFIG.dat', 'RUN_AT_STARTUP:TRUE;' + #13#10, False) then
                     begin
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -39,7 +39,7 @@ Name: "desktopicon"; Description: "{cm:CreateDesktopIcon}"; GroupDescription: "{
 [Files]
 Source: "..\ka-lite\dist\ka-lite-static-*.zip"; DestDir: "{app}\ka-lite"
 Source: "..\khan_assessment.zip"; DestDir: "{app}"
-Source: "..\ka-lite\scripts\*.bat"; DestDir: "{app}\ka-lite\scripts\"
+Source: "..\scripts\*.bat"; DestDir: "{app}\ka-lite\scripts\"
 Source: "..\gui-packed\KA Lite.exe"; DestDir: "{app}"; Flags: ignoreversion
 Source: "..\gui-packed\guitools.vbs"; DestDir: "{app}"; Flags: ignoreversion
 Source: "..\gui-packed\images\logo48.ico"; DestDir: "{app}\images"; Flags: ignoreversion
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -386,6 +386,8 @@ var
   restoreKaliteFolder: integer;
   restoreContentFolder: integer;
   informationBoxFlagged: boolean;
+  kaliteSchtaskCmd: string;
+  windowsVersion: TWindowsVersion;
 
 begin
     if CurStep = ssInstall then
@@ -444,7 +446,18 @@ begin
       
             if StartupPage.SelectedValueIndex = 0 then
             begin
-                if Exec(ExpandConstant('{cmd}'), ExpandConstant('/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"'), '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
+                GetWindowsVersionEx(windowsVersion);
+
+                { Windows 5.1 is XP, and 5.2 is Server 2003/64-bit XP. See https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832(v=vs.85).aspx }
+                if (windowsVersion.Major = 5) and (windowsVersion.Minor <= 2) then
+                begin
+                    kaliteSchtaskCmd := ExpandConstant('/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart"');
+                end
+                else begin
+                    kaliteSchtaskCmd := ExpandConstant('/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"');
+                end;
+
+                if Exec(ExpandConstant('{cmd}'), kaliteSchtaskCmd, '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
                 begin
                     if Not SaveStringToFile(ExpandConstant('{app}')+'\CONFIG.dat', 'RUN_AT_STARTUP:TRUE;' + #13#10, False) then
                     begin
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -444,7 +444,7 @@ begin
       
             if StartupPage.SelectedValueIndex = 0 then
             begin
-                if Exec(ExpandConstant('{cmd}'), '/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"', '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
+                if Exec(ExpandConstant('{cmd}'), ExpandConstant('/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"'), '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
                 begin
                     if Not SaveStringToFile(ExpandConstant('{app}')+'\CONFIG.dat', 'RUN_AT_STARTUP:TRUE;' + #13#10, False) then
                     begin
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -451,7 +451,7 @@ begin
                 { Windows 5.1 is XP, and 5.2 is Server 2003/64-bit XP. See https://msdn.microsoft.com/en-us/library/windows/desktop/ms724832(v=vs.85).aspx }
                 if (windowsVersion.Major = 5) and (windowsVersion.Minor <= 2) then
                 begin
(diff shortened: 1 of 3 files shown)


MCGallaspy:
Check out [this build](http://dungeon.learningequality.org:8085/browse/KL-KALITE016-WIN-87) to test it out!

MCGallaspy:
@radinamatic let's give it a shot

radinamatic:
Gottit! Downloading...

radinamatic:
No luck, @MCGallaspy, at least on Windows 8 (testing 7 now):

- I still get "Failed to create config file/option" alerts
- Scheduled task is created correctly, and at least according to system, it is executed correctly too, but KA Lite does not run at startup.
 (screenshot coming)

radinamatic:
Windows XP:

- No "Failed to create config file/option" alerts
- Scheduled task is created correctly, but KA Lite does not run at startup. 

wxp

radinamatic:
However, if I activate all three options and restart the machine, both KA Lite and server are started! :)

MCGallaspy:
> I still get "Failed to create config file/option" alerts

That's expected -- this only addresses the "start at boot" option

radinamatic:
Windows 7 yields the same as result as 8:  Scheduled task is created correctly, but KA Lite does not run at startup.

(screenshots of some task properties coming)

radinamatic:
However, as noted before for XP, if I activate the "Run KA Lite when the user logs in" and restart, that task gives us the desired result (KA Lite is run at startup). Could this be a user problem?

MCGallaspy:
> if I activate the "Run KA Lite when the user logs in" and restart, that task gives us the desired result (KA Lite is run at startup). 

Does it truly run when the system starts? Or just when that user logs in? I'm specifically trying to debug the "at system start" option. If the "when user logs in" options works ok, try logging in as a different user and seeing if it starts -- I do not expect it to.

radinamatic:
You are probably right, I'll create another user (this one came with VM image by default), and check that out.

MCGallaspy:
I'm attempting to install & replicate, but in the meantime it could be informative to run the script which the scheduled task is configured to run, to see if there is any error output.

radinamatic:
Running (double clicking) `kalite.bat` opens and immediately closes the command prompt window, so to be able to see the output, I made a copy of the `.bat` file, edited it in Notepad adding `@pause` at the end and run it again. This is a output:

selection_120

MCGallaspy:
yeah, `kalite.bat` is meant to be run from the command line

MCGallaspy:
Believe I know why, @radinamatic if you can confirm. During installation, we create the scheduled task using the script from this PR, which relies on expanding the variable %KALITE_SCRIPT_DIR%. But that variable is not expanded because the environment will not be updated until the setup exe exits. The result is a misconfigured scheduled task:

virtualbox_ie9 - win7_18_12_2015_14_04_20

Note that the full path to the script is missing. Radina, can you confirm that post-installation, having checked the "start at system startup" option, that the scheduled task is misconfigured the same way it is here?

radinamatic:
My above "Actions" screenshot from W7 seems identical @MCGallaspy.
I restarted the W8 VM and this is what's inside the KA Lite task:

ie11 - win8 1 py 3 5 1 first running - oracle vm virtualbox_121

I'm gonna revert and do another fresh install, just in case I meddled with the above one.

radinamatic:
After fresh install on W8, it seems the same as in your example @MCGallaspy:

ie11 - win8 1 pre ka lite running - oracle vm virtualbox_124

MCGallaspy:
Yeah, I see. I just submitted a patch, should be the latest build on bamboo. Let's try installing that one "fresh".

radinamatic:
Ok, tomorrow I'll be testing build # 90 on all the machines, and report back! @MCGallaspy

MCGallaspy:
TODO: once this is merged, I need to modify the repos used on the build server.

MCGallaspy:
Looking more closely at your screenshots, there's some bizarre options for the tasks (like only running when on AC power). Seems to be the default, however... Will address those here as well.

radinamatic:
Win XP tested today with build `#91`, still no luck with start on boot... :(

xp-build91

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -444,7 +444,7 @@ begin
       
             if StartupPage.SelectedValueIndex = 0 then
             begin
-                if ShellExec('open','guitools.vbs','4', ExpandConstant('{app}'), SW_HIDE, ewWaitUntilTerminated, StartupCode) then
+                if Exec(ExpandConstant('{cmd}'), '/C "schtasks /create /tn "KALite" /tr "\"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat\" start" /sc onstart /ru SYSTEM"', '', SW_HIDE, ewWaitUntilTerminated, StartupCode) then
```
radinamatic: I was trying to understand this line of code and run into this, maybe it could be of help...?

http://stackoverflow.com/questions/3636055/how-to-modify-the-path-variable-definitely-through-the-command-line-in-windows

radinamatic:
Ok, I've been at this for a while now, trying various options why the task may have `(0x1)` as "Last Run Result" as listed **[here](https://social.technet.microsoft.com/Forums/windowsserver/en-US/57c65328-d8c8-4e13-9b51-e017ae6c298c/task-scheduler-returns-0x1-code-when-run-whehter-user-is-logged-on-or-not-options-selected?forum=winserverManagement)**, when I realized I needed to clarify something, mainly to myself.

If the user selects `Run the server at system startup`, that would imply that, if the scheduled task is working correctly, `python` process is activated on startup, but not `KA Lite.exe`. This has been failing on all the systems so far. 

I know it doesn't help much since you are trying to work out the solution with the `KALITE_SCRIPT_DIR`, but I did manage to start the `python` process by putting the local path `C:\Python27\Scripts` in both "Program/Script" and "Start in (optional)" options of the "Actions" tab, the second one without the quotes:

Bad news is that even with this `python` process running and listening to the correct port, I still cannot open KA Lite in the browser: it reports "Connecting..." and "Waiting for 127.0..0.1..."

(screenshot coming)

Could it have something to do with the "Last Run Result" now being `The task is currently running (0x41301)` and **not** being `successful`?

Another issue is that once user starts `KA Lite.exe` itself from the desktop shortcut (if they choose to add it), the second option `Options > Run KA Lite at system startup` seems be activated by default, but it is not working, namely, `KA Lite.exe` does not  run on startup:

boot3

This possibly deserves a separate issue.

radinamatic:
Correction, server did start at the end, but it took good 10-15 minutes to do so... :P
I'll retest this tomorrow on a another VM with more RAM, maybe that was the issue. 

So, maybe the next step is to figure out how to pull that directory from the registry and put it in the "Start in (optional)" option too? I tried on my end, but given that it's a registry path (with blank spaces in between on top of everything), I couldn't figure out the right approach. 
@MCGallaspy

radinamatic:
The above test case (putting the local path `C:\Python27\Scripts` in both "Program/Script" and "Start in (optional)" options of the "Actions" tab, the second one without the quotes), worked in Windows 8.1 too.

(screenshot coming)

I wasn't able to test it on Windows XP because the Task Scheduler is a bit more "primitive", and is picky in allowing changes.

radinamatic:
Regarding XP, though, seems that it's the only one that fills in the "Start in:" option of the "Actions" tab by default, from the script that you have in the installer, while Windows 7 & 8 have only the "Program/Script" option after the installation. So, a part from figuring out why that reg:HKLM fails to grab the value of  `KALITE_SCRIPT_DIR` env var (and I did check, **it is in the registry**), try forcing the "Start in:" option, because it is mentioned on several places that I checked as a potential cause of scheduled task fail.

MCGallaspy:
Looks like you've been busy, so I'm addressing comment by comment:

Re: https://github.com/learningequality/installers/pull/322#issuecomment-166383187

The values enclosed in `{` & `}` should be expanded by IS -- but it seems they weren't. I'll double check. After installation finishes, what is the value of that registry key? You can check using `regedit`, and the path is given, i.e. the root key is HKLM, then System, etc, and you can follow it until you get to Environment.

windows/installer-source/KaliteSetupScript.iss
MCGallaspy: The values in the curly braces should have been replaced by IS with the given registry value -- but it seems it wasn't.

radinamatic:
If I recall well, the value of the registry key was the (correct) Python scripts path on that machine.

I'll fire up the VM and recheck.

MCGallaspy:
I read up, looks like the registry value is set properly -- is it substituted correctly in the scheduled task on other version of Windows? In other words, does it fail to substitute only on XP?

You also mentioned that you added C:\Python27\Scripts to your PATH, and the scheduled task ran properly -- did you edit the scheduled task prior to changing the PATH value? In other words, there are four cases. Can you fill in the values in the table below?


|      | bad schtask | good schtask          |
| ---- | --------- | ----------- |
|    **Scripts dir in PATH**   |   ?      | ? .|
|   **Scripts dir *not* in PATH**   | Does not work     |  ?     |

radinamatic:
**Yes**, registry value is set correctly on all versions (XP, 7, 8), and **No**, it's not substituted correctly in scheduled task, so **Yes**, task fails on all the versions.

I did not edit/add Python to the path, installer does that correctly, I just edited the task itself, substituting the unexpanded `{...}` in the task with `C:\Python27\Scripts`.

(screenshot coming) 

It might be in one of the links that I posted above, but there might have been a suggestion somewhere that in order for it to expand correctly, you must set the reg value type as `REG_EXPAND_SZ` instead of `REG_SZ`.

radinamatic:
Addendum to the above comment: `C:\Python27\Scripts` is added to the PATH **IF** the user ticks the appropriate checkbox during the installation. If they do not, or Python 2 is the sole one installed, it's not the problem, as it is taken by default. This only becomes a problem when there is a Python 3 in the mix too, and its path **is** added (see #296).

MCGallaspy:
Okay, this latest commit should expand the constant if you picked the "Start at boot" option initially. We can test it out from this build: http://dungeon.learningequality.org:8085/browse/KL-KALITE016-WIN-93

radinamatic:
Good news @MCGallaspy, so far it seems to be working on Windows 7, raising the Python process correctly... :)

I'll test on rest of the versions and report back!

radinamatic:
We are not out of the woods just yet... :(

Build # 93 Worked flawlessly on Windows 7 & 8, but not on XP. Scheduled task seems OK (`C:\Python27\Scripts` is expanded correctly), but Python process is not raised, and server never starts. `kalite.bat` works when run from Terminal. 

I've been at this for a while, can't figure out why. If you have any ideas, yell!

MCGallaspy:
It may be the user that the scheduled task is run as (SYSTEM) -- if it works when you run as yourself, can you try opening a cmd.exe prompt as SYSTEM (this option should be available from the "Run as..." context menu when you right click) then running kalite.bat to see what happens?

MCGallaspy:
Yeah, after experimenting with it I think the solution is to run the scheduled task as a different user on XP only -- looking in to it.

radinamatic:
If I try to open cmd.exe prompt from the "Run as...", I don't get to choose SYSTEM, it's not available as an option. There is only "Administrator" user alongside the default VM user, and if I choose it, it demands a password that I'm unaware of (VM only offers password for the default user).

Scheduled tasks log shows the following:
```
"KALite.job" (kalite.bat)
	Finished 12/30/2015 8:29:37 AM
	Result: The task completed with an exit code of (1).
```

`1 or 0x1` exit code implies incorrect or unknown function called, if I recall well. Could it be the that the argument `start` is not passed correctly? In XP it does not have a separate field in the task properties window as for 7 or 8.

MCGallaspy:
I'm not sure what the error code precisely means without looking at the output of the script, which I'm not really sure how to do. However, I found that when the task is run as a different user (not SYSTEM) then it works without error.

radinamatic:
True, if I change the scheduled task (or make a new one) with `Run as:` set for the default VM user, task is executed correctly, python process is raised and server starts on boot no prob...

So does that mean that for XP we'll have to configure the installer to run `kalite.bat` task set for the logged-in user?

MCGallaspy:
@radinamatic okay, this should start the task in Windows XP. For XP only, the scheduled task is set to run as the installing user -- in this case it should be the admin, since the installer is run with elevated privileges. I'm afraid this might make upgrading on XP a little tricky, since it may mean *moving* files from one location to another. But for later versions of Windows, this should introduce no change. If this works well for a regular install, then I'll merge this and we can begin testing upgrade scenarios, with an emphasis on XP.

radinamatic:
Ok, waiting on build # 94 to test this @MCGallaspy.

radinamatic:
Build #94 seems to be problematic only on XP, on Windows 7 installer finishes correctly, and starts the server Ok on restart.

MCGallaspy:
Yeah, on XP it prompts the user for a password.

MCGallaspy:
I'm disabling the start at boot option from the task tray program on XP. Shortly I'll remove the option from the installer.

MCGallaspy:
@radinamatic I'm merging this so we can start working on #321.

PR #339, opened 2016-01-11
Tooltip help isn't too user friendly

### LINKED ISSUE #338: Tooltip help isn't too user friendly
I'm referring to this:

screen shot 2016-01-08 at 08 44 28

My expectation is that if I click that button a help window would pop up. Instead nothing is happening, and the only way I even knew that the tooltip containing help text would appear was that someone informed me. Else, I'd think this button isn't working.

Strongly suggest to have a pop up window instead.

amodia:
@aronasorman & @cpauya  this fixes #338 

I show an alert window to display the help text.

v1


cpauya:
Merging this now until we come up with the proper help window for the application.

PR #343, opened 2016-01-25
Feature/338 tooltip help isnt too user friendly part 2

### LINKED ISSUE #338: Tooltip help isn't too user friendly

cpauya:
Hi @aronasorman - this fixes #338.  This is based on the initial work by @amodia.

We now use a [Popover](https://developer.apple.com/library/prerelease/mac/documentation/UserExperience/Conceptual/OSXHIGuidelines/ControlsView.html#//apple_ref/doc/uid/20000957-CH52-SW2) instead of the alert dialog when the "?" button is clicked by the user.  This is more user-friendly than the tooltip and more appropriate than the alert.  It also shows the context of the field where the "?" belongs.

Here's a screenshot:
screenshot 2016-01-25 11 39 45

### DIFF
--- a/osx/KA-Lite/KA-Lite.xcodeproj/project.pbxproj
+++ b/osx/KA-Lite/KA-Lite.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 46;
+	objectVersion = 47;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


cpauya:
Merging this now so Bamboo can create a new OS X installer for testing.

PR #344, opened 2016-01-25
Feature/340 restart is no longer required after successfull installation

### LINKED ISSUE #340: A restart is required after installation.
**Branch:** Develop (OS X installer)
**Current behavior:** After a successful installation the installer wizard prompt you to restart.
After successful installation, the `KALITE_PYTHON` environment variable is running on the sudo context. We need to restart the machine to load the .plist file which sets the `KALITE_PYTHON` environment variable in the user context so that it will be accessible in `KA-Lite.app`.

**Expected behavior:** After a successful installation the `org.learningequality.kalite.plist` file should be loaded via script so that the `KALITE_PYTHON` environment variable is set.  Remove the restart required prompt in our installer wizard.

**Checklist:**
- [x] Load the `org.learningequality.kalite.plist` via script.
- [x] Remove restart required.

cpauya:
Hi @aronasorman this fixes #340.  This is based on the initial branch created by @arceduardvincent.

To prevent a restart after installation, we now set the `KALITE_PYTHON` environment variable under the user account doing the installation.  

This also works when doing an upgrade installation where we set the value of the env var to the new location of the Pyrun executable (`/Users/Shared/ka-lite/pyrun-2.7/bin/pyrun`).

We now do `syslog` messages during installation so that users can trace their install process, here's a screenshot:
screenshot 2016-01-25 17 44 14

### DIFF
--- a/osx/KA-Lite-Packages/KA-Lite.pkgproj
+++ b/osx/KA-Lite-Packages/KA-Lite.pkgproj
@@ -644,7 +644,7 @@
 				<key>AUTHENTICATION</key>
 				<false/>
 				<key>CONCLUSION_ACTION</key>
-				<integer>2</integer>
+				<integer>0</integer>
 				<key>IDENTIFIER</key>
 				<string>org.learningequality.kalite</string>
 				<key>LOCATION</key>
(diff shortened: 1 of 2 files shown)


cpauya:
Merging this now so Bamboo can create a new OS X installer for testing.

PR #348, opened 2016-01-27
[Windows] Fix new scheduled task behavior

### LINKED ISSUE #345: Run server on system startup not working after upgrade to 0.16 (build 104)
Run server on system startup not working after upgrade to 0.16 (build 104)

##### Upgrade scenario

0.15.1 > 0.16 (build 104) on Windows 7
Server was starting correctly on boot with 0.15.1 installation, independently from which user (admin or standard) logged in. 

##### Expected Behavior

If the user upgrades to 0.16 and enables `Run KA Lite on system startup` among Options in task-tray, KA Lite server should start on boot irregardless of which user logs in. 

##### Current Behavior

KALite scheduled task fails with `...The specified service does not exist. (0x800704DD)` error:

##### Screenshots

ie11 - win7 0 16 -104 running - oracle vm virtualbox_218

ie11 - win7 0 16 -104 running - oracle vm virtualbox_219

##### Steps to reproduce

1. Install 0.15.1.
2. Upgrade to 0.16.
3. Start KA Lite.exe.
4. Activate "Run KA Lite on system startup" in KA Lite task-tray Options
5. Restart system.
6. Check the Task Scheduler.

MCGallaspy:
Fixes #345. Commit messages below:

### Enabling system start task prompts user for password
```text
This is because Windows will refuse to start the KA Lite server
in the background *before* the user logs in if the password is
not provided. Note that passing the /np (no password) flag will
cause the task to *attempt* to start, but it will fail because
"no logon server could process the request".
```

### 	When upgrading to 0.16.0, remove old "KALite" scheduled task
```
It's run as the SYSTEM user, which has been an error since
0.13.0 -- see #345. Delete it so we can start fresh.
```

### DIFF
--- a/windows/scripts/add_systemstart_task.bat
+++ b/windows/scripts/add_systemstart_task.bat
@@ -9,5 +9,5 @@ if %WIN_VERSION% LEQ "5.2" (
     echo This feature is unavailable on this version of Windows.
     pause
 ) else (
-    schtasks /create /tn "KALite" /tr "\"%KALITE_SCRIPT_DIR%\kalite.bat\" start" /sc onstart /ru %USERNAME% /f
+    schtasks /create /tn "KALite" /tr "\"%KALITE_SCRIPT_DIR%\kalite.bat\" start" /sc onstart /ru %USERNAME% /rp /f
 )


MCGallaspy:
@radinamatic hold off on testing until I reconfigure the build server.

radinamatic:
Roger!

MCGallaspy:
I just started build 107, should be good to go. See if it addresses the issues you mentioned in the referenced issue (#345).

radinamatic:
After the first test on upgrade 0.15 > 0.16 (with build # 110), the old SYSTEM user task was still there... Re-testing now to confirm.

radinamatic:
Tested the upgrade 0.14.2 > 0.16 (build # 110) on W7, and again the old SYSTEM user task was still present in Task Scheduler & `C:\Windows\System32\Tasks`.

radinamatic:
In both upgrade scenarios, since the old task is still there, I delete it and install the new from task-tray options. This new one is setup to run only if the user is logged in, but it fails to raise python processes:

task
 
If I change the task to run whether user is logged in or not, **and** add the password, it finally manages to start the server. How is this option reflected in production settings? Are the users required to input their password? Every time?

MCGallaspy:
Okay, seemingly the old task is not removed correctly. I'm looking in to that. However, regardless of whether the old task is there, I would expect creating the new task using the task-tray program would prompt you for a password -- are you saying that doesn't happen? If not, what are the results of running the following command in a cmd.exe shell? `schtasks /create /tn "KALite" /tr "\"%KALITE_SCRIPT_DIR%\kalite.bat\" start" /sc onstart /ru %USERNAME% /rp /f`

MCGallaspy:
Ah, yeah, it's not building from this branch. Whoops! That would be why the error isn't resolved.

MCGallaspy:
Let's try again starting from build 112.

radinamatic:
I thought that might be the case... 

I might have also figured out another thing that was confusing me for a while. It's related to tasks, but I better put it in a new issue for clarity sake.

radinamatic:
Actually I'm gonna hold on reporting that until I confirm it on build # 112.

MCGallaspy:
Try build 113 now.

radinamatic:
We're on the right track @MCGallaspy! 

Old task is eliminated correctly and the new one installs and is able to start the server upon system restart... :+1: 

Tested on **0.15.1 > 0.16** upgrade scenario on W7, I'll repeat to confirm with other base installs, but I'm confident it's going to be OK. 

I'll open issues for two new bug/enhancement proposals tangential to this, but I think we can move on and close the original issues for this PR.

radinamatic:
Opened 2 new issues that became evident and confirmed during this testing scenario: #351 & #352.

radinamatic:
On another hand, build 114 works equally well on **0.14.2 > 0.16** upgrade scenario on W7. 

I think you can close this one as you will @MCGallaspy! :wink:

PR #350, opened 2016-01-29
Delete old KALite start-at-boot task when upgrading to 0.16

### LINKED ISSUE #345: Run server on system startup not working after upgrade to 0.16 (build 104)

radinamatic:
Remove the old scheduled task run as SYSTEM user, since from now it should be run as current user.
Should fix #345.

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -241,6 +241,7 @@ end;
 procedure HandleUpgrade(targetPath : String);
 var
     prevVerStr : String;
+    retCode: Integer;
 begin
     prevVerStr := GetPreviousVersion();
     if (CompareStr('{#TargetVersion}', prevVerStr) >= 0) and not (prevVerStr = '') then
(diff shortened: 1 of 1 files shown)


MCGallaspy:
I can't configure the build server to use your branch -- it needs your credentials apparently :P

This looks correct -- have you tested it?

PR #362, opened 2016-02-21
New templates for issues and PRs

radinamatic:
Second take on this. Definitely less confusing now, thanks @benjaoming!!! :+1: 

Now that I think of it, do test & dependencies items in PR template make sense here, on installers? Or should I leave them out?


benjaoming:
Merging this right away, suggesting that changes are more than welcome as we go along :)

PR #367, opened 2016-02-26
0.16 debian changes

benjaoming:
## Summary

All changes for 0.16 prereleases tracked here.


benjaoming:
From and including 4f74860, everything is being released as 0.16.2-ubuntu2 -- **INCLUDING A MAJOR PROBLEM FIX**, namly that assessment items bundled in were from 0.15.. not the database itself but the static files. Upgrading will automatically fix this, no matter if you're on 0.16.1 or 0.16.2-ubuntu1.

benjaoming:
Tested this by means of upgrading from 0.15.1 to 0.16.2-ubuntu2.

benjaoming:
Merging because we need a cleaner way of tracking changes, so in the future, I'll open up a PR for every new ubuntu pkg release.

PR #374, opened 2016-03-09
Unpack content pack with correct option

MCGallaspy:
## Summary

Using the --foreground option prevents database
access, which is an error if `setup` has not been
run (which is the case at install time).


MCGallaspy:
Seems no one is the habit of reviewing my PRs for the Windows installer, so I'll go ahead and merge.

PR #420, opened 2016-09-19
Removal of *pyc in debian installer

### LINKED ISSUE #418: Debian installer does not remove pyc files in /usr/share/kalite
## Summary

Upgrades seem to be broken because of the change in requests version. The underlying problem is that there are lingering *pyc files.

This is because of a simple typo.

## Branch or installer

 - Installer: Debian

Related:
https://github.com/learningequality/ka-lite/issues/5277

benjaoming:
## Summary

Fixes #418


benjaoming:
This will be released as `0.16.9-ubuntu2` and replace the current installers.

PR #421, opened 2016-09-20
Do not depend on possibly problematic systemd library

benjaoming:
## Summary

Not released yet

On a vanilla Raspbian installation, I found the system service causing this issue, it was unseen on previous installations of Raspbian Jessie, so I can't say for sure what has gone wrong.

```
pi@raspberrypi:~ $ sudo /etc/init.d/ka-lite restart
[....] Restarting ka-lite (via systemctl): ka-lite.serviceFailed to restart ka-lite.service: Unit ka-lite.service failed to load: No such file or directory. failed! 
```

## TODO

> PR is considered WIP (work in progress), until all TODOs are marked.

- [ ] Added entry to CHANGELOG.rst

### DIFF
--- a/debian/debian/init-service
+++ b/debian/debian/init-service
@@ -2,7 +2,7 @@
 
 # Author: Foundation for Learning Equality
 #
-# /etc/init.d/kalite
+# /etc/init.d/ka-lite
 
 ### BEGIN INIT INFO
 # Provides:          kalite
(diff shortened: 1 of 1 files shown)


benjaoming:
Seems to be related to this issue:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=825913

Trying to upgrade the RPi firstly, but otherwise the true fix would be to fetch and include the auto-generated ka-lite.service....

benjaoming:
FYI @aronasorman

benjaoming:
Even though this works, it seems sort of sad to disable all the system features that will bring colorful terminal outputs and `[ OK ]` messages.. going to create a real systemd startup script instead...

PR #427, opened 2016-11-16
Updating windows installer for 0.17

### LINKED ISSUE #424: Updating Windows installer for 0.17
## Summary
- [ ] `compileymltojson` is no longer a command
- [ ] Docs are built into static-libraries/docs with the `make docs` target, this should be invoked or copied into the `bamboo-build.sh` script?
- [ ] Test that the `kalite.bat` executable still works.
- [ ] Use the new auto-discovered content pack (optional)
## Version
- Installer: Windows
- Version: 0.17b1

mrpau-richard:
## Summary

- Use the KA Lite static .whl file.
- Update the Python version from 2.7.9 to 2.7.11.
- Set KALITE_PYTHON environment variable after the installation.

## Issues encountered.
Having `multiprocessing issue` when using the python version 2.7.10
temp-1

## Issues addressed

Fixes #424 

##

Hi @radinamatic can you test this windows installer for 0.17 release? I built it in my local for now because the build server is still down. Here's the [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMTW9zdU5vT3ZYeVU).

##

/cc @cpauya @benjaoming


radinamatic:
Downloading! 

I'll try it on Windows 7, unless you have other preferences...?  @mrpau-richard

radinamatic:
@mrpau-richard Your installer worked perfectly on W7, no errors or issues whatsoever during the installation, nor in major functions.

All caveats regarding the contentpacks still the same as for the `deb` installer I tested yesterday.

benjaoming:
@mrpau-richard great work! Does it mean we have to add Python 2.7.11 as a requirement for Windows environments? We could add it as a release note with an instruction on how to find your python version...

mrpau-richard:
@benjaoming Yeah, We need to add it in the release notes.

benjaoming:
@mrpau-richard I managed to identify the upstream Python issue, it's true that multiprocessing is broken in < 2.7.11

https://bugs.python.org/issue10128

benjaoming:
Allow me to rant :)

Python 2.7.11 is only 11 months old.

For starters, if someone is already deploying a new version of KA Lite, they should also have access to an updated version of Python. It's within reach of anyone installing KA Lite on Windows to also install Python.

Requiring users to run a newer version of Python also shouldn't do harm, but could come with other benefits as lots of bugs are fixed in each of the 2.7 patch releases.

We may be making our own work much more difficult if we decide to allow users to stay with old versions of Python. Supporting old OS'es is one thing that's already quite a load :)

So in that sense, I would say that we should stick to the changes you've made.. keep on building the Windows installer with wheels and instead require users to read the release notes, where we'll mention that Windows requires a Python update.

benjaoming:
@mrpau-richard - why not bundle Python 2.7.12 since it's the latest release?

mrpau-richard:
I haven't tested it within python 2.7.12. I will test it tomorrow if I have no issue encountered then we will use it.

mrpau-richard:
Hi @benjaoming & @radinamatic we are now using the python version 2.7.12 in KA Lite windows installer. Here's the [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMYWxGejRzSUhxVk0) for testing.

radinamatic:
@mrpau-richard Tested the installer on Windows XP this morning, and everything seems a-OK! 👍 

One thing I would suggest is to rephrase the wording, since it seems to imply that the Python installation is somehow optional, and not compulsory - users cannot simply skip installation of Python and expect KA Lite to work:

virtualbox_wxp_18_11_2016

Suggested text for the above alert:

```
KA Lite requires Python to be installed on your system. 
This wizard will proceed to install Python 2.7.12, and then continue with KA Lite installation. 
```

The proper buttons should be `Continue` and `Cancel`, and if `Cancel` is selected, there should be another alert with the following:

```
Are you sure you want to cancel the installation of Python 2.7.12? 
If you select "Yes", this installer will close and KA Lite will not be installed on your system. 
```

mrpau-richard:
Hi @radinamatic done fixing the text for the alert box. Here's [the installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMUG9xbGt6WVRvWlE)

radinamatic:
@mrpau-richard Text looks OK, thanks! Two new issues however:

### Issue 1:

When user decides not to install Python/KA Lite, and chooses `Yes` here:

virtualbox_wxp_21_11_2016_cancel

then this Runtime error appears on XP:

virtualbox_wxp_21_11_2016_cancel2

Same on Windows 7:

virtualbox_windows 7_21_11_2016_cancel


### Issue 2:

If the user selects `No` on the alert step/screenshot above, I think the installer should go back to the previous logical step, namely this one, and wait for the user to select `OK`:

virtualbox_wxp_21_11_2016_accept

Instead, the process remains running in the back and eventually starts the Python installer **without** the user's explicit acceptance.

mrpau-richard:
Hi @radinamatic I fix the issue you commented above. Here's the [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMUWN3aXN6Qk9EUHM) for another testing.

radinamatic:
@mrpau-richard Awesome, no more runtime errors! 👍 

I'll give it another cursory glance, and see if I can debug the contentpacks a bit more!

mrpau-richard:
Hi @radina Can you do a final testing on KA Lite windows installer for 0.17? To merge this PR.
You can download the latest installer at the [dungeon server.](http://dungeon.learningequality.org/browse/KL-MT0-59) 

I already tested it, on these operating systems Windows 7 32 bit and Windows 8.1 64 bit.

radinamatic:
@mrpau-richard Tested the build 59 on both W7 & W8.1 64bits and  as far as the fixes in this PR are concerned, merge away! 👍

mrpau-eduard:
Hi @benjaoming can you take a look on this `PR` it seems it is ready for merging.

benjaoming:
Skimmed through the code, still looks good, thanks for the tip @mrpau-eduard :)

PR #435, opened 2017-01-23
KA Lite server taking so long to start on El Capitán

mrpau-richard:
## Summary

This fixes KA Lite server taking so long to start on El Capitán when using the latest English content packs with subs.

##

Hi @radinamatic can you test again, the latest OS X [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMVmhwTHpmOHVQVHM) with the updated English content pack. 

/cc @radinamatic  @cpauya @mrpau-eduard

### DIFF
--- a/osx/KA-Lite/KA-Lite/AppDelegate.h
+++ b/osx/KA-Lite/KA-Lite/AppDelegate.h
@@ -31,10 +31,10 @@
 @property (strong, nonatomic) NSStatusItem *openItem;
 @property (unsafe_unretained) IBOutlet NSTextView *taskLogs;
 
-// It will increment if port 8008 is not available.
-@property signed int portAlertCounter;
-// It will increment when startFunction() will execute.
-@property signed int kaliteStartCounter;
+// It will set as YES if port 8008 is not available.
+@property BOOL isPortAlert;
+// It will set as YES when startFunction() will execute.
+@property BOOL iskaliteStart;
 
 @property BOOL isLoaded;
 @property BOOL autoStartOnLoad;
--- a/osx/KA-Lite/KA-Lite/AppDelegate.h
+++ b/osx/KA-Lite/KA-Lite/AppDelegate.h
@@ -30,8 +30,12 @@
 @property (strong, nonatomic) NSStatusItem *stopItem;
 @property (strong, nonatomic) NSStatusItem *openItem;
 @property (unsafe_unretained) IBOutlet NSTextView *taskLogs;
-@property signed int kaliteProcessCounter;
+
+// It will increment if port 8008 is not available.
 @property signed int portAlertCounter;
+// It will increment when startFunction() will execute.
+@property signed int kaliteStartCounter;
+
 @property BOOL isLoaded;
 @property BOOL autoStartOnLoad;
 @property NSString *version;
--- a/osx/KA-Lite/KA-Lite/AppDelegate.m
+++ b/osx/KA-Lite/KA-Lite/AppDelegate.m
@@ -206,7 +206,6 @@ BOOL isKaliteCommand(NSString *command) {
     task = [[NSTask alloc] init];
     
     if (isKaliteCommand(command)) {
-        self.kaliteProcessCounter += 1;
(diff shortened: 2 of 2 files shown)


radinamatic:
Finished testing this latest installer on Sierra, but I keep having problems with running El Capitan VM at all: basically it keeps logging me off with a slightest change made, and makes it impossible to test effectively.. 😭  

Anyways, the installer works OK on Sierra, English subs load out of the box, tested Spanish and `pt-PT` language download from the UI, no sign of the topic tree error you encountered on the Windows installer... 👍 

One question, why does it still say `0.17.0dev0` in the footer? Is that different for every OS? The `deb` one I tested on Friday said (correctly) `0.17.0b5`...?

virtualbox_sierra_23_01_2017_13_07_56

@mrpau-richard 
@mrpau-eduard

mrpau-richard:
@cpauya can you also review this PR.

osx/KA-Lite/KA-Lite/AppDelegate.m
```diff
@@ -270,26 +269,34 @@ - (enum kaliteStatus)checkRunTask:(NSNotification *)aNotification{
     if (!isKaliteCommand(lastCommandArg)) {
         if (status == statusOkRunning) {
             self.portAlertCounter += 1;
```
cpauya: This is confusing me, if we are just checking if a port is used, then why use increment/decrement when a simple Boolean value will do?  Any particular reason for this?

osx/KA-Lite/KA-Lite/AppDelegate.m
```diff
@@ -270,26 +269,34 @@ - (enum kaliteStatus)checkRunTask:(NSNotification *)aNotification{
...
+                [self setNewStatus:statusStopped];
                 [self showStatus:statusStopped];
+                return self.status;
+            } else {
+                if (self.portAlertCounter == 1) {
+                    self.portAlertCounter -= 1;
+                }
+                // This will run `kalite start` command after checking the port 8008.
+                if (self.kaliteStartCounter == 1) {
+                    self.kaliteStartCounter = statusOkRunning;
```
cpauya: Why did you assign an `enum` value here instead of an `int`?  Shouldn't this be an increment or decrement?  Basically this assigns a zero value to `kaliteStartCounter`, right?

osx/KA-Lite/KA-Lite/AppDelegate.h
```diff
@@ -30,8 +30,12 @@
 @property (strong, nonatomic) NSStatusItem *stopItem;
 @property (strong, nonatomic) NSStatusItem *openItem;
 @property (unsafe_unretained) IBOutlet NSTextView *taskLogs;
-@property signed int kaliteProcessCounter;
+
+// It will increment if port 8008 is not available.
 @property signed int portAlertCounter;
+// It will increment when startFunction() will execute.
+@property signed int kaliteStartCounter;
```
cpauya: Does this need to be an `int`?  How about using a `Boolean` instead as I saw below that not much increment/decrement/computation is happening on this variable.

osx/KA-Lite/KA-Lite/AppDelegate.m
```diff
@@ -1048,7 +1041,7 @@ - (void)discardPreferences {
 - (void)startKaliteTimer {
     // TODO(cpauya): Use initWithFireDate of NSTimer instance.
     // TODO(amodia): Check if kalite environment variables change.
-    [NSTimer scheduledTimerWithTimeInterval:5.0
+    [NSTimer scheduledTimerWithTimeInterval:08.0
```
cpauya: Why did we increase the delay to 8 seconds?  Is that why we removed the "delay" codes above?

mrpau-richard: I increase the delay because sometimes the application will freeze, when starting KA Lite server.

mrpau-richard:
@cpauya this fixes the KA Lite server taking so long to start on El Capitán  when using the latest English content packs with subs.

PR #437, opened 2017-02-02
Fix not loading English content pack with subtitles using the KA Lite windows application.

mrpau-richard:
Hi @radinamatic Can you test this using this [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMaUwtdXJnSC1jdmc).

/cc @cpauya  @mrpau-eduard

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -470,6 +470,7 @@ var
 begin
     { Used to have more responsibility, but we delegated those to the app itself! }
     { Unpacks the English content pack. }
+    Exec(ExpandConstant('{cmd}'), '/S /C "' + ExpandConstant('"{reg:HKLM\System\CurrentControlSet\Control\Session Manager\Environment,KALITE_SCRIPT_DIR}\kalite.bat"') + ' manage syncdb --noinput"', ExpandConstant('{app}'), SW_HIDE, ewWaitUntilTerminated, retCode);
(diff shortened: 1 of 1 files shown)


benjaoming:
@mrpau-richard `manage retrievecontentpack local en en.zip --foreground` <- This is no longer relevant if you install `en.zip` to the well-known location before running `setup` / starting KA Lite. See: https://github.com/learningequality/ka-lite/pull/5402

radinamatic:
@mrpau-richard Topic tree problem seems fixed (:+1:) as it loads with bare installation... However I kept getting the C.O.R.S. error I [reported previously](https://github.com/learningequality/ka-lite-content-packs/issues/58). 

virtualbox_ie11 - win8 1_02_02_2017_12_19_11

virtualbox_ie11 - win8 1_02_02_2017_12_50_42

benjaoming:
@radinamatic that issue isn't an installer issue, though. You should be able to reproduce it with the source distribution (`pip install ka-lite --pre`) - and most likely also see it in 0.16 ?

mrpau-richard:
Hi @benjaoming As we discuss on our last sync up we cannot extract the English content pack on the `kalite start` command, because it will take more time to start the KA Lite server. We will implement it later when we have a logging UI in KA Lite windows application.

benjaoming:
@mrpau-richard sounds good! There's already an issue #438 for this.

As for the other error that @radinamatic is reporting, it's already reported elsewhere and is largely solved.

radinamatic:
Yep, re-tested this installer on a fresh W8.1 VM, no CORS error, and the latest `en` contentpack installed in 5 minutes! 👍 

virtualbox_ie11 - win8 1_08_02_2017_12_20_51

virtualbox_ie11 - win8 1_08_02_2017_12_21_21

PR #439, opened 2017-02-09
Stop the KA Lite server during upgrade.

### LINKED ISSUE #351: Stop the server running in background prior to upgrade
Stop the server running in background prior to upgrade to 0.16
##### Upgrade scenario

0.13 through 0.15 > 0.16 on W7 (build # 114)
##### Current Behavior

In KA Lite versions 0.13 through 0.15 users had the option to run the server on system startup with the scheduled task. Since that task was run as the SYSTEM user, current user of KA Lite is not even able to stop the server. During the upgrade process, server is still running background and that is a potential problem for two reasons:
1. Could be causing the issue #347 - [error described here](https://github.com/learningequality/installers/pull/335#issuecomment-171485593). We haven't been able to reproduce it reliably, but could be still related to the server running in the background during upgrade.
2. Bad user experience: Once the upgrade process is finished, the old scheduled task is deleted, but server is still running (1-2 python processes are active) and current user cannot stop it. User is also unable to start the server on their own since task tray program registers the server that is running and displays only the `Stop Server` option as available, but it cannot stop the python process run as SYSTEM user. Task-tray program also displays the `Load in browser` option as available, but if user selects it, it will only open a blank page at server's IP. 

At this point user is left with the only option of restarting the system to regain control of the KA Lite server. We could include that requirement in the installer, or even better: stop the server prior to starting the upgrade process. 
##### Steps to reproduce
1. Install 0.15.1.
2. Upgrade to 0.16.
3. Start KA Lite.exe.
4. Activate "Load in browser" in KA Lite task-tray Options.

mrpau-richard:
## Summary

> Stop the KA Lite server during upgrade.

##

Hi @radinamatic can you verify the fix using this [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMamRrcUlsbDg1SzQ).

## Issues addressed

Fixes #351 

/cc @benjaoming @cpauya @radinamatic

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -423,19 +423,11 @@ begin
              exit;
         end
     else
-      if(MsgBox('Warning! All python process will be closed in order to upgrade your Python.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idCancel) then
-          begin
-          HandlePythonSetup();
-          exit;
-          end
-      else begin
-          Exec(ExpandConstant('taskkill.exe'), '/f /im ' + '"python.exe"', '', SW_HIDE, ewWaitUntilTerminated, installPythonErrorCode);
-          ExtractTemporaryFile('python-2.7.12.msi');
-          ExtractTemporaryFile('python-2.7.12.amd64.msi');
-          ExtractTemporaryFile('python-exe.bat');
-          ShellExec('open', ExpandConstant('{tmp}')+'\python-exe.bat', '', '', SW_HIDE, ewWaitUntilTerminated, installPythonErrorCode);
+        ExtractTemporaryFile('python-2.7.12.msi');
+        ExtractTemporaryFile('python-2.7.12.amd64.msi');
+        ExtractTemporaryFile('python-exe.bat');
+        ShellExec('open', ExpandConstant('{tmp}')+'\python-exe.bat', '', '', SW_HIDE, ewWaitUntilTerminated, installPythonErrorCode);
 
-      end
 end;
 
 function InitializeSetup(): Boolean;
@@ -449,9 +441,19 @@ begin
   
     ShellExec('open', 'taskkill.exe', '/F /T /im "KA Lite.exe"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode)
     ShellExec('open', 'tskill.exe', ' "KA Lite"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode);
-
     RegDeleteValue(HKCU, 'SOFTWARE\Microsoft\Windows\CurrentVersion\Run', ExpandConstant('{#MyAppName}'));
-   
+    if(MsgBox('Warning! All python process will be closed in order to install KA Lite.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idCancel) then
+    begin
+        if(MsgBox('Are you sure you want to cancel the installation of KA Lite?' + #13#10 + 'If you select Yes, this installer will close and KA Lite will not be installed on your system.', mbConfirmation, MB_YESNO) = idYes) then
+          begin
+            forceCancel := True;
+            ExitProcess(1);
+          end
+          else begin
+            InitializeSetup()
+          end
+    end
+    Exec(ExpandConstant('taskkill.exe'), '/f /im ' + '"python.exe"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode);
     if ShellExec('open', 'python.exe','-c "import sys; (sys.version_info >= (2, 7, 11,) and sys.version_info < (3,) and sys.exit(0)) or sys.exit(1)"', '', SW_HIDE, ewWaitUntilTerminated, PythonVersionCodeCheck) then
     begin
         if PythonVersionCodeCheck = 1 then
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -442,17 +442,7 @@ begin
     ShellExec('open', 'taskkill.exe', '/F /T /im "KA Lite.exe"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode)
     ShellExec('open', 'tskill.exe', ' "KA Lite"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode);
     RegDeleteValue(HKCU, 'SOFTWARE\Microsoft\Windows\CurrentVersion\Run', ExpandConstant('{#MyAppName}'));
-    if(MsgBox('Warning! All python process will be closed in order to install KA Lite.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idCancel) then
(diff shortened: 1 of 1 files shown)


benjaoming:
If this can be verified, it can make it for 0.17 :+1:

radinamatic:
At first I was not exactly sure in what way this fix and the new installer are supposed to "stop" the server running in the background... 🤔 

I tested on W7 & W8.1 with previous KA Lite versions installed, and what happens is that, since 0.17 installer carries the Python version 2.7.12, this one checks the previously installed Python version before it even gets to KA Lite installation. So in case KA Lite server is running in the background, Python installation will itself instruct the user to close the `python` process so it can proceed:

virtualbox_ie11 - win8 1_09_02_2017_11_14_31

Basically, in above scenarios there was no way for me to test if the new KA Lite installer actually stops the server running in the background, because it's already been stopped by the 2.7.12 Python installation, which always comes first. At this point I'm not even sure if this feature is needed now, as in all upgrade-to-0.17 scenarios the Python updater will always require the user to kill the background `python` process, effectively stopping the KA Lite server. 

So just to be thorough, I did do a "control" test with installing this newest `0.17b5` over the previous `0.17b4` release, and there I did manage to observe that the background `python` process is, in fact, killed by the KA Lite installer (since the Python was already version 2.7.12), prior to proceeding with `0.17b5` installation.

To sum it up, the fix works 👍 , but will be of the real use only when users start upgrading from 0.17 to 0.18 (if we ever get there), since during the upgrade from previous versions, the Python installer itself will do the job. Could be useful in cases when user starts the upgrade to 0.17, but stops the KA Lite installer after upgrading Python, and re-starts it in another moment.

benjaoming:
@radinamatic thanks for testing! I guess we can't include a Python upgrade in all installers, so the fix is valuable still because it will prevent the installer from crashing during minor upgrades :)

mrpau-richard:
Hi @radinamatic & @benjaoming I created, a confirmation message to kill all Python process before running the latest python installer.

Here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMNC1NV2paUDM1QUk).

radinamatic:
@mrpau-richard I was able to see the confirmation message, however it enters in a sort of endless loop and the user ends up being unable to continue and update Python:

1. First problem is that the installer kills both the task-tray KA Lite app, and the 2 python processes even **BEFORE** asking for the confirmation from the user. This should occur after the user has accepted that python processes can be closed.

2. After the confirmation message appears, it looks like the same process is started and closed again, and there's no way around... :(

w8b


Windows 8, latest 0.17.0 installer.

benjaoming:
Let's get 0.17.0 released and update the installer with this fix after. There's no problem releasing a second 0.17.0 installer for Windows with this update immediately after 0.17.0 ships.

radinamatic:
@benjaoming Agreed, I did not consider this any kind of release blocker... :wink:

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -423,19 +423,19 @@ begin
              exit;
         end
     else
-      if(MsgBox('Warning! All python process will be close in order to upgrade your Python.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idYes) then
-        begin  
+      if(MsgBox('Warning! All python process will be close in order to upgrade your Python.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idCancel) then
```
radinamatic: Change `close` into `closed`.

mrpau-richard:
Hi @radinamatic Here's the new [KA Lite Windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMV2RGd1B5cjdOQzA) that fixes the issue in the confirmation message.

benjaoming:
@mrpau-richard Was this released for 0.17.0?

mrpau-richard:
@benjaoming this was not released  in 0.17.x.

benjaoming:
Alright, let's see if @radinamatic would like to do a final test - in case it works, perhaps we should upload a fixed Windows installer instead of waiting for 0.17.1?

radinamatic:
Ok, installing the [most recent Windows 0.17 installer that I'm aware of from here](https://drive.google.com/drive/folders/0B5Tyi-tRKOFMblJYY0RIM0U1TzA), over an older one I had laying around, did not let me see the alert message. Even though server was running when I opened the installer, both Python processes and the `KA Lite.exe` task tray program were closed without any input from the user. Could it be that the Python installer bundled into KA Lite installer did that on it's own?

Testing the installation of 0.17.0 over 0.16.9, will report back!

radinamatic:
Even on 0.17.0 over 0.16.9 installation, Python process is closed **before** the user is consulted, but at this point I'm not sure if we can control this:

w7e

mrpau-richard:
@radinamatic I've moved the confirmation message before the installation of Python and KA Lite. Here the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMNWgyY0M0elRfdU0) for testing.

capture

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -437,9 +441,19 @@ begin
   
     ShellExec('open', 'taskkill.exe', '/F /T /im "KA Lite.exe"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode)
     ShellExec('open', 'tskill.exe', ' "KA Lite"', '', SW_HIDE, ewWaitUntilTerminated, killErrorCode);
-
     RegDeleteValue(HKCU, 'SOFTWARE\Microsoft\Windows\CurrentVersion\Run', ExpandConstant('{#MyAppName}'));
-   
+    if(MsgBox('Warning! All python process will be closed in order to install KA Lite.' + #13#10 + #13#10 + 'Click OK to continue.', mbConfirmation, MB_OKCANCEL) = idCancel) then
```
radinamatic: Change into `All Python processes will be closed...`.

mrpau-richard: @radinamatic I updated the confirmation message as you suggested. Here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMY25aV0NVUDRWV2c) to test it

radinamatic:
Tested the latest installer, and I'm 😕 as to what exactly changed...

As you ca see, the Python processes in the background were closed **before** the user had a chance to give the OK, right after the UAC permission is given:

w7g

benjaoming:
ping @mrpau-richard :)

benjaoming:
Closing this because of branch clean-up.

Please reopen for `master` @mrpau-richard

benjaoming:
ping @mrpau-richard - would be nice to have this wrapped up, do you follow @radinamatic 's comments and can you reproduce?

mrpau-richard:
@radinamatic I can't reproduced your comment's above. with this [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMNWgyY0M0elRfdU0).

When I run the installer the python process still exist until I click the `ok` button at the dialog message.
gif-temp

radinamatic:
Let's go with this for now, the fix works as intended, and I'm only too picky about the issues 99% of the users wouldn't even care about... 😛

mrpau-richard:
Merging this. Thanks @radinamatic

PR #442, opened 2017-03-07
Allow user to select location of the old content folder during upgrade installation on windows

### LINKED ISSUE #341: Allow user to select location of the old content folder during upgrade installation on Windows
@MCGallaspy is implementing the steps to find content folder of the previous installation and move it (if necessary) to the new location during the upgrade to 0.16 (PR  #335). In case installer is unable to find the location of the content folder for the previous version automatically, the following step would be to allow users to browse and select the old-version content folder themselves. Suggested wording:

```
KA Lite Setup is unable to locate the content folder of your previous installation 
in order to perform the upgrade to current version. Please select the ".kalite/content/" 
folder on your computer and press the button "Continue" to finish the upgrade process. 
```

mrpau-richard:
## Summary

> I created a dialog box to allow user to select the location of ./kalite/content folder if it's not exist during the upgrade.

## Issues addressed

Fixes #341 

/cc @cpauya @radinamatic


mrpau-richard:
Hi @radinamatic can you verify the fix using this [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMbFlBMTVZdmZpNDQ)

benjaoming:
Ping @radinamatic :)

benjaoming:
Ping @radinamatic again :)

radinamatic:
On this now!

And btw I'm seeing the `CORS header missing` error again, while preparing 0.16.9 version for this update scenario:

virtualbox_windows 7_27_04_2017_13_40_25

radinamatic:
Ok, tried the [Windows installer linked above](https://drive.google.com/open?id=0B5Tyi-tRKOFMbFlBMTVZdmZpNDQ), and the implementation of the feature seems to work correctly:

1. Detects the prior installation and asks if user wishes to upgrade:
virtualbox_windows 7_27_04_2017_14_25_55

2. Is unable to find the previous `.kalite/content` folder and allows the user to select it:
virtualbox_windows 7_27_04_2017_14_26_11

3. Copies the content into the standard location in `Users/<username>/.kalite/content/` folder:
virtualbox_windows 7_27_04_2017_14_26_42

However, I'm not sure if this feature only copies the `/content/` folder into the new (standard) location, and not the rest of the previous `/.kalite/` folder, which would leave the database and locales back. That results in new issues:

1. User is asked to register the new admin account.
2. Device needs to be registered again.
3. Errors on viewing Facilities and Languages:

virtualbox_windows 7_27_04_2017_14_58_16

virtualbox_windows 7_27_04_2017_14_52_21

benjaoming:
@radinamatic that doesn't sound perfect? Copying contents can be very expensive - I thought the point was to either change the `CONTENT_ROOT` setting or to create a symbolic link (not sure how that works on Windows).

rtibbles:
I think changing the CONTENT_ROOT setting is probably preferable.

radinamatic:
@benjaoming Loosing the previously installed locales is definitely not the end of the world, as contentpacks can be re-downloaded, but I bet nobody would look favorably too loosing the database...
 
I agree that copying the content may not be the most efficient solution, and if changing the  `CONTENT_ROOT` is better, let's go for it. I'm just saying that if we go the "copying" route, let's copy the database together with the `content` folder, because that will almost certainly be the expected outcome in case of an upgrade.

benjaoming:
I'm confused now.

Are we talking about allowing the user to select the `CONTENT_ROOT` (folder with videos) or the user data directory root (also known as `~/.kalite`, the one with database and content packs.. and by default also the one with videos) ?

radinamatic:
I am not sure which exact scenario inspired the original issue that prompted the work on this fix, it was long time ago... My guess is that we envisioned the possibility that the previous version stored content in some "non recommended" location, and the decision was to use the `user/.kalite/` folder for both videos and databases during the upgrade. 

The point is that my original suggestion and this implementation of just copying the `content` folder is not enough, as it may leave the existing database behind. The solution would be to check for existing database also, and if so, copy it too in the `user/.kalite/` folder (we don't have the `DATABASE_ROOT`, right?). Or don't go the copying route at all, but think of other solution how to make available the old  database after the upgrade.

benjaoming:
Closing this because of branch clean-up.

Please reopen for `master` @mrpau-richard

benjaoming:
@mrpau-richard do you have an answer to @radinamatic 's question above?

> However, I'm not sure if this feature only copies the /content/ folder into the new (standard) location, and not the rest of the previous /.kalite/ folder

mrpau-richard:
Hi @radinamatic I agree that setting the CONTENT_ROOT environment variable is more efficient  than copying it to ./kalite folder. About the old database I think we can create an option to allow the user to import or not the existing database from the CONTENT_ROOT path.

radinamatic:
@mrpau-richard I agree, let's go with offering the user the option to import the old database during the upgrade! :+1:

mrpau-richard:
@radinamatic here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMeWJqbTFzZ25ITXc) that allow's the user to import an existing database from the selected content folder.

radinamatic:
Ok, so to test this I moved the `./kalite` folder outside the user folder, to the root of the C: drive.

The new installer correctly detected the previous installation and reported that no content folder could be found. When prompted I selected the whole `./kalite` folder on the root of the C: drive, and the installation continued:



However, once the installation finished and I started the server, it required me to create the admin user again, which was the first sign something has gone awry. So I opened the `data.sqlite` file, which was 3Kb smaller than original one, and there was no sign of the learner user I previously created:



The original db did have one learner user:



@mrpau-richard Not sure which part has gone wrong, but the original database was not preserved during the upgrade process.

benjaoming:
@radinamatic the database isn't part of the content folder, this feature would only allow using videos stored in a specific location. Can you confirm that this is the intention @mrpau-richard ?

radinamatic:
Understood, @benjaoming, I was under the impression that if I selected the whole `./kalite` folder, it will be backed up, including the database. 

So, what exactly is the user supposed to do in order to import the old database? @mrpau-richard

mrpau-richard:
@radinamatic In order to import the old database it must be copied to `./kalite/database` folder.

radinamatic:
@mrpau-richard Any reason we are not copying the `./kalite/database` folder together with `./kalite/content`. I was under the impression that was what we agreed upon few weeks ago...

> create an option to allow the user to import or not the existing database.

mrpau-richard:
@radinamatic sorry, I wasn't aware how you test this. When the installer prompted you to select content folder it most be pointed at `.kalite/content` not in `.kalite` folder it will also detect existing database to import it. I will also try to fix the conflict after this [PR](https://github.com/learningequality/ka-lite-installers/pull/457) merged.

benjaoming:
@radinamatic @mrpau-richard I think we should only aim for having the content folder imported. This is a far more common use case - we can see about the database folder another time.

radinamatic:
@mrpau-richard My bad, I'll retest and point to `.kalite/content` folder this time.

@benjaoming 
We can, of course, deal with the user database import in another PR, if it's easier to implement separately, but I fail to see why would somebody want only the content and not the all the users and their activity log preserved during the upgrade... 😕 

What if the facility had setup several classes/groups with many users? If the database is not imported during the upgrade, coach/admin will have to create them from the scratch, all the data for coach tools will be lost, and all the learners will loose their previously earned points. Am I missing something? Don't we want to preserve that...?

benjaoming:
> but I fail to see why would somebody want only the content and not the all the users and their activity log preserved during the upgrade...

You are right about this, but I think it's already preserved?

Actually, I thought the change would also affect fresh installations, in which case I thought it was nice because users may have `content` sitting on a different drive. For instance, because they are replicating an installation on a new device or have downloaded the contents separately via torrent.

mrpau-richard:
@radinamatic here's the [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMTXlpQ3FOeG9fY0E) to test this PR again. 
I also created another [issue](https://github.com/learningequality/ka-lite-installers/issues/459) to allow the user to import there existing database during upgrade.

@benjaoming I agree, That we allow the user to import there 'content' during fresh installation. Are we going to implement it on this PR?

benjaoming:
> @benjaoming I agree, That we allow the user to import their 'content' during fresh installation. Are we going to implement it on this PR?

Maybe it's better to say "select" than "import" (as in the title of this PR) because it denotes that files are moved or copied.

If you think it's easy to move this logic so it triggers both on fresh installs and upgrades, that'd be great! :+1:

radinamatic:
@benjaoming 
> You are right about this, but I think it's already preserved?

In update/upgrade scenarios from the more recent versions (where everything is already in `user/.kalite/` folder) it is indeed, but I believe here we are exploring the options of what to preserve when the user had the old version of KA Lite in a location which the installer does not recognize, and is not able to preserve the old content and database. The fix here takes care only of the `content` folder, and I'm trying to "squeeze in" the preservation of the user database, too... 😉   

> Actually, I thought the change would also affect fresh installations, in which case I thought it was nice because users may have content sitting on a different drive. For instance, because they are replicating an installation on a new device or have downloaded the contents separately via torrent.

All completely true and I support us thinking of these scenarios, I'm just saying we should check if the user database is also sitting somewhere else than the default folder. I see @mrpau-richard already opened another issue to deal with that 👍 👏 , so I'll finish drilling about it here... 😛

mrpau-richard:
@radinamatic here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMYVhGcjdyMkVHTUE) that's allow the user to select content folder during fresh and upgrade installation.

radinamatic:
I tested this and it seems to work OK, that is it ask the user to select the `.kalite/content/` folder on fresh install.

However, I'm worried if this option results confusing for the new and less savvy users... :thinking: 
Could we add this phrase at the bottom of the alert window for fresh install?

`If this is the first time you are installing KA Lite, or you don't need to import any previous content at this time, press Cancel.   `

benjaoming:
Suggestion to Radina's text:

> You can download contents after the installation. If you don't have a folder with downloaded contents, press Cancel.

The dialogue has the following potential purposes:

1. Selecting contents from a previous installation
1. Selecting contents readily downloaded on a non-default location (USB drive etc)
1. Selecting a content folder on a different drive because the default drive doesn't have enough space (and then downloading contents after the installation)

mrpau-richard:
@radinamatic & @benjaoming I updated the alert message. you can grab the installer [here](https://drive.google.com/open?id=0B5Tyi-tRKOFMdFIxOUhFbTV0Vmc)

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -242,14 +242,42 @@ begin
...
+    retCode: Integer;
+begin
+    userPath := ExpandConstant('{%USERPROFILE}')
+    contentPath := ExpandConstant('{%CONTENT_ROOT}')
+    systemKaliteDir := ExpandConstant(userPath + '\.kalite')
+    userKaliteContent := ExpandConstant(systemKaliteDir + '\content');
+    if Not DirExists(userKaliteContent) then
+    begin
+     if Not DirExists(contentPath) then
+        begin
```
benjaoming: I this condition you are saying: 

> If there is no user content folder AND there is no previously declared `CONTENT_ROOT` environment

But this only matches that there already

1. was a previous installation AND
1. the previous installation does not have a content folder in the default location
1. whatever content folder was selected previously doesn't exist

I think for a user to go through the installation and then being asked questions, and then if you try to re-install and *not* get asked the same questions might be problematic? If they want to correct a mistake for instance, they have to go through the whole uninstallation (in which I'm not sure if removal of environment registry is guaranteed?)

My suggestion would be to completely remove the condition and instead display the path to the user.

In the future, we might also display a count of free space available on the drive.

windows/installer-source/KaliteSetupScript.iss
```diff
@@ -242,14 +242,42 @@ begin
...
+            if MsgBox('You can also download contents after the installation. If you don`t have a folder with downloaded contents, press Cancel.', mbInformation,  MB_YESNO or MB_DEFBUTTON1) = IDYES then
+            begin 
+               if BrowseForFolder('Please select the .kalite/content/ folder', userPath, False) then
```
benjaoming: Is it normal that the default location of `BrowseForFolder` is *not* the same as what's already selected (`userPath` instead of `userKaliteContent`) ?

benjaoming: The text:

> Please select the .kalite/content/ folder

Assumes that the user knows that `.kalite/content` is just a suggestion.. this sounds a lot like an instruction to select that precise folder.

Suggesting:

> Please select the content folder where you wish downloaded videos to be stored.

benjaoming (changes requested):

benjaoming:
@mrpau-richard I was commenting in the context of @radinamatic 's comment, so not thinking that my suggestion should stand by itself :)

> However, I'm worried if this option results confusing for the new and less savvy users... :thinking:
Could we **add this phrase** at the bottom of the alert window for fresh install?

My suggestion in full would be to replace this text:

`'KA Lite Setup is unable to locate the content folder of your previous installation in order to perform the upgrade to current version. Please select the .kalite/content/ folder on your computer and press the button OK to finish the upgrade process.'`

With:

`'KA Lite Setup cannot find any video contents in the default location (' + contentPath + '). You can download contents after the installation or select a location with previously downloaded contents. If you don't have a folder with downloaded contents, press Cancel. To change the default location of downloaded contents, press OK.'`

`contentPath` should default to `userKaliteContent` when it's unset.

windows/installer-source/KaliteSetupScript.iss
mrpau-richard: Hi @benjaoming if we remove this condition we also need to remove this `'KA Lite Setup cannot find any video contents in the default location (' + contentPath + ').` 

 In this sentence  
`'KA Lite Setup cannot find any video contents in the default location (' + contentPath + '). You can download contents after the installation or select a location with previously downloaded contents. If you don't have a folder with downloaded contents, press Cancel. To change the default location of downloaded contents, press OK.'`

We also can't say if there's a content folder but it's empty, unless we check the file size of it.

This will always ask the user set there content path during installation but it's optional.

benjaoming: Oh you're right, the text only really fits in that case.

> This will always ask the user set there content path during installation but it's optional.

IMO it's fine to ask the user as long, as *if they are confused*, they will find it easy to just press cancel.

Let's try generalizing it some more:

```
'KA Lite Setup will look for video contents in the folder ' + contentPath + '. You can download contents to the selected folder after the installation completes - or - select a folder with already downloaded contents.\n\nIf you don't have a folder with downloaded contents or want to use the suggested folder, press Cancel.\n\nTo change the default folder for downloaded contents, press OK.'
```

Is that better?

benjaoming: I don't know if `\n` works in the message box?

mrpau-richard:
@benjaoming & @radinamatic I updated the dialog message as you guys suggested. You can grab the installer [here](https://drive.google.com/open?id=0B5Tyi-tRKOFMZGpUTUNBSVVNaG8) for testing.
screen shot 2017-07-13 at 4 40 30 pm

radinamatic:
I have been re-reading our conversation on this for a while now, have a grim impression that we've lost focus on what we are trying to achieve (probably good part is my fault too :disappointed:), and I think that we need to pause and rewind.

Not sure if it is the formatting issue for the text inside installer dialogs, or there might be some copy/paste errors between various versions of the message here, but I find this wording to be very unclear and potentially confusing to the user, especially if it is their first encounter with KA Lite.

## TL;DR version

After spending quite some time on writing all that below, I realized it will be much clearer if I made a drawing... :stuck_out_tongue: 

selecting content folder for import during the ka lite installation - 0 17 2 feature proposal

### Yeah, I want to read

1. If we are to proceed with implementing this complex feature that covers several potential purposes, I would insist we add a conditional first dialog for users who don't need it, or are installing for the first time: 

> If this is the first time you are installing KA Lite, or you don't need to import any previous local content at this time, skip this step to continue the installation.
> 
> Note: You will be able to download videos from the KA Lite management interface.

This dialog should have **two** buttons at the bottom:

- `Import content now` (first/left)
- `Skip importing content` (second/right; in focus by default, so user just needs to press `Enter` to continue)

2. If there is the default content folder from previous installation: 

> KA Lite Installer found videos in the default location (' + contentPath + '). 
> 
> Do you want to import this content from the previous installation, or you want to explore locations on other drives?

This dialog should have **three** buttons at the bottom:

- `Back` (first/left; for users who ended here but changed their mind) 
- `Explore other drives` (second/middle)
- `Import from default location` (third/right; in focus by default, so user just needs to press `Enter` to continue)

If the user chooses `Explore other drives` installer should open the standard Windows `Open folder` dialog, and `Import from default location` should directly proceed to import content. **Both options should include a confirmation message at the end**, before continuing with the installation. 

This should take care of all three proposed purposes:
- [X] Selecting contents from a previous installation
- [X] Selecting contents readily downloaded on a non-default location (USB drive etc)
- [X] Selecting a content folder on a different drive because the default drive doesn't have enough space (and then downloading contents after the installation)

benjaoming:
@radinamatic Just sharing my 2 cents here: Changing the content folder (to another drive for instance) after the installation and running KA Lite for the first time is quite annoying as it requires:

1) Editing `~/.kalite/settings.py` or doing something with symlinks (which afaik doesn't work on Windows anyways)
2) Copying and quite possibly merging to different folders (as the original content folder will contain all of the content pack stuff... assessment items, subtitle files etc).

I'm speaking as a user of the Python/Debian stuff... it doesn't matter if you have a helpful wizard or not. Actually, not having it is quite unhelpful, as it will unpack the existing pre-seeded content pack automatically and then you have to stop KA Lite and create a new location for the content folder, symlink stuff (or change settings), copying over all the files of the older content folder.. which is a bit undocumented.

benjaoming:
@radinamatic Just sharing my 2 cents here: Changing the content folder (to another drive for instance) after the installation and running KA Lite for the first time is quite annoying as it requires:

1) Editing `~/.kalite/settings.py` or doing something with symlinks (which afaik doesn't work on Windows anyways)
2) Copying and quite possibly merging to different folders (as the original content folder will contain all of the content pack stuff... assessment items, subtitle files etc). You should put all videos (downloaded from for instance torrent) into a folder `content/khan`, and I don't know how well this stuff is really documented.

I'm speaking as a user of the Python/Debian stuff... it doesn't matter if you have a helpful wizard or not. Actually, not having it is quite unhelpful, as it will unpack the existing pre-seeded content pack automatically and then you have to stop KA Lite and possibly create a new location for the content folder, symlink stuff (or change settings), copying over all the files of the older content folder..

But I like that we try to ask users the right questions during installation, and I think either way - the current simple dialogue (not sure if the displayed default path is correct) or a more intelligent flow that you suggests - will be a great improvement.

radinamatic:
@benjaoming I misunderstood the last option you  were proposing initially - `Selecting a content folder on a different drive because the default drive doesn't have enough space (and then downloading contents after the installation)` - if I'm understanding it correctly now, it means that user chooses to use another drive as the location for `./kalite/content/` folder, right? 

In that case that step (letting the user select the **new** location where the videos will be) should come before choosing where to search the content to import from, correct? This means that the above workflow needs some adjustments I'll get right to that!

benjaoming:
> In that case that step (letting the user select the new location where the videos will be) should come before choosing where to search the content to import from, correct?

Yes, this is true - thanks for observing this, as I was perhaps a bit in a hurry :) But yes, I'm trying to cater for users that are BOTH 1) installing for the first time and 2) upgrading and 3) re-installing to make another choice for the content folder (as this is a dialogue you do not find in KA Lite itself after installing)

benjaoming:
Note: The goal of closing this issue is to make something that's reusable in Kolibri

radinamatic:
Closing as wontfix

PR #443, opened 2017-03-19
Fixing systemd

benjaoming:
Lacking a mandatory section, obviously because the auto-generated file, I copied was broken.

Systemd is becoming the standard on most major distros, so we better be up to speed. Also this for consideration: https://twitter.com/systemdsucks

## Summary

Tested in Ubuntu 16.04

## Issues addressed

https://github.com/learningequality/ka-lite/issues/5423

## TODO

> PR is considered WIP (work in progress), until all TODOs are marked.

- [x] Added/updated the documentation.
- [x] Updated installer's local changelog
- [x] Verified test builds

### DIFF
--- a/debian/debian/changelog
+++ b/debian/debian/changelog
@@ -1,3 +1,9 @@
+ka-lite-source (0.17.0-0ubuntu2) trusty; urgency=medium
+
+  * Rebuild with systemd fix
+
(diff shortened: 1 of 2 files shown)


benjaoming:
Verified in Ubuntu 16.04 and Raspian Jessie - works!

    systemctl start/stop/restart ka-lite  #systemd notation
    service ka-lite start/stop/restart  # old notation

PR #448, opened 2017-04-26
Fix hardcoded KA Lite version at mac OS X installer.

### LINKED ISSUE #426: OSX installer claims it's 0.16.0 not 0.16.9
## Summary

> The opening screen shows ￼
> version 0.16.0 rather than 0.16.9 
## Version
- Installer: OSX
- Version: 0.16.9
## Traceback or server.log snippet

```
None, and not relevant.
```
## Screenshots

Attached.
## How to reproduce
1. Run KA-Lite-0.16.9.pkg
2. Check the text in the opening screen.
   screen shot 2016-10-04 at 17 21 49

Please can you ensure the version number is consistent in the installer, the app and the installer.

mrpau-richard:
## Summary

> This will fix the hardcoded KA Lite version at the `.pkg`  installer.

## Issues addressed

Fixes #426

### DIFF
--- a/osx/build.sh
+++ b/osx/build.sh
@@ -373,9 +373,9 @@ fi
 echo "Updating KA Lite installer version"
 KA_LITE_VERSION="$(kalite --version)"
 DOCS_PATH="$SCRIPTPATH/KA-Lite-Packages/docs"
-sed -i '' "13s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/INTRODUCTION.rst.rtfd/TXT.rtf"
-sed -i '' "12s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/SUMMARY.rtfd/TXT.rtf"
-sed -i '' "12s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/README.rst.rtf"
+sed -i '' "13s/KA_LITE_VERSION/$KA_LITE_VERSION/g" "$DOCS_PATH/INTRODUCTION.rst.rtfd/TXT.rtf" && \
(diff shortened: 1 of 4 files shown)


osx/build.sh
```diff
@@ -368,6 +369,17 @@ if [ $? -ne 0 ]; then
     exit 1
 fi
 
+((STEP++))
+echo "Updating KA Lite installer version"
+KA_LITE_VERSION="$(kalite --version)"
+DOCS_PATH="$SCRIPTPATH/KA-Lite-Packages/docs"
+sed -i '' "13s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/INTRODUCTION.rst.rtfd/TXT.rtf"
```
benjaoming: Here, you are replacing `0.17.0` - it would be better to change the files to have a placeholder like ``{{ KALITE_VERSION }}`` and replace this instead?

osx/build.sh
```diff
@@ -368,6 +369,17 @@ if [ $? -ne 0 ]; then
...
+sed -i '' "12s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/SUMMARY.rtfd/TXT.rtf"
+sed -i '' "12s/0.17.0/$KA_LITE_VERSION/g" "$DOCS_PATH/README.rst.rtf"
+if [ $? -ne 0 ]; then
```
benjaoming: This only tests the output of the previous command failed.

You can do the following to test all of it:

```bash
sed ... &&
sed ... &&
sed ... ||
(echo "failed to update installer version" && exit -1)
```

mrpau-richard:
@benjaoming Thanks for reviewing this.  I've done fixing all your comments.

Issue #444, opened 2017-03-30
Tests for Zesty Zapus


benjaoming:
## Summary

Since 16.10, systemd has been both for system and user sessions (startup applications etc). No more upstart.

We've fixed #440 and as of now, released it in ka-lite-proposed

I think we should test our installers for 17.04 already, so we know if we're good.

@radinamatic - would you be able to do this?

radinamatic:
Tested on Ubuntu 16.10, with installer [ka-lite-bundle_0.17.0-0ubuntu2_all.deb](https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite-proposed/+files/ka-lite-bundle_0.17.0-0ubuntu2_all.deb)

However, server is not starting at reboot:

![virtualbox_ubuntu 16 10_10_04_2017_14_21_48](https://cloud.githubusercontent.com/assets/1457929/24862548/54c190e6-1dfd-11e7-9a1e-49fbeba5098b.png)

![virtualbox_ubuntu 16 10_10_04_2017_14_22_05](https://cloud.githubusercontent.com/assets/1457929/24862569/60b895c0-1dfd-11e7-95cf-6d0b8be0ec56.png)

Tried checking `kalite.service` as suggested in [#5423](https://github.com/learningequality/ka-lite/issues/5423):

![virtualbox_ubuntu 16 10_10_04_2017_14_38_11](https://cloud.githubusercontent.com/assets/1457929/24862583/6c4e1db0-1dfd-11e7-9c3c-dcfcbc4d746d.png)

benjaoming:
@radinamatic I'll get 16.10 running on a virtualbox and see what that's about.

benjaoming:
Seems everything is fine with the latest 0.17.0-0ubuntu3 - which incidentally also contained a systemd fix, see #443

radinamatic:
I tried again on Ubuntu 16.10 (latest with all the updates), with installer `ka-lite-bundle_0.17.0-0ubuntu3_all.deb`, but the server is still not starting at reboot.

![virtualbox_ubuntu 16 10_03_05_2017_15_57_13](https://cloud.githubusercontent.com/assets/1457929/25664732/ced7a6ce-301b-11e7-8c6f-0ae58a648606.png)

![virtualbox_ubuntu 16 10_03_05_2017_16_12_03](https://cloud.githubusercontent.com/assets/1457929/25664743/d613bed2-301b-11e7-8f2f-b912eb62b05c.png)

[...]

Issue #440, opened 2017-02-20
Start-up script `/etc/init.d/ka-lite` does not get added to default runlevels


benjaoming:
## Summary

I haven't verified this issue on Debian, but on the latest Raspbian based on Debian Jessie, the start-up script does not run at bootup.

Running `sudo update-rc.d ka-lite defaults` has no effects, apparently the old script is a noop.

## Workaround

Non at the moment...

## Version

Any .deb installer on Raspbian.

## Traceback

```
root@raspberrypi:~# /etc/rc3.d/S02ka-lite status
● S02ka-lite.service
   Loaded: not-found (Reason: No such file or directory)
   Active: inactive (dead)
```

benjaoming:
Confirmed on Ubuntu 16.10

While systemd script is running `systemctl start ka-lite` just fine, the links created by `update-rc.d` do not seem to start.

The startup links are created correctly:

```
# ...
lrwxrwxrwx 1 root root 21 May  2 17:03 S01grub-common -> ../init.d/grub-common
lrwxrwxrwx 1 root root 20 May  2 17:03 S01irqbalance -> ../init.d/irqbalance
lrwxrwxrwx 1 root root 17 May  3 12:08 S01ka-lite -> ../init.d/ka-lite
```

But again, something is wrong with the systemd compatibility:

```
root@kalite-VirtualBox:~# /etc/rc3.d/S01ka-lite status
Unit S01ka-lite.service could not be found.
```

Running the script directly from `/etc/init.d/ka-lite` seems to work, so it could be that somehow our script or systemd doesn't like the symlink:

```
root@kalite-VirtualBox:~# /etc/init.d/ka-lite status
● ka-lite.service - LSB: ka-lite daemon, a local Khan Academy content server
   Loaded: loaded (/etc/init.d/ka-lite; static; vendor preset: enabled)
   Active: active (exited) since Wed 2017-05-03 14:31:29 CEST; 3min 7s ago

May 03 14:31:25 kalite-VirtualBox systemd[1]: Starting LSB: ka-lite daemon, ....
May 03 14:31:25 kalite-VirtualBox su[2722]: Successful su for kalite by root
May 03 14:31:25 kalite-VirtualBox su[2722]: + ??? root:kalite
May 03 14:31:25 kalite-VirtualBox su[2722]: pam_unix(su:session): session op...)
May 03 14:31:29 kalite-VirtualBox ka-lite[2714]: Running 'kalite start' as d...)
[...]

PR #450, opened 2017-05-05
Systemd startup

benjaoming:
## Summary

The general issue is that while the [Debian Wiki](https://wiki.debian.org/Teams/pkg-systemd/Integration) claims this:

> update-rc.d(8) should enable/disable systemd services in addition to creating the sysvinit links. update-rc.d is usually called from maintscripts, but enable/disable is typically used by the system administrator. 

...we can't run the `/etc/rcX.d/K01ka-lite` scripts that are added as links, they give a systemd error, and I'm not going to pursue it any further after finding that `systemctl enable ka-lite` works.

 * Removes confusing and unused files
 * Adds systemctl enable for systemd

## Issues addressed

#440 
#444


benjaoming:
The output from Travis tests confirms that it's still running on Ubuntu Trusty, i.e. a system w/o systemd.

benjaoming:
This is released in a matter of hours on the ka-lite-proposed PPA: https://launchpad.net/~learningequality/+archive/ubuntu/ka-lite-proposed/+packages

PR #457, opened 2017-06-08
Create a menu to show KA Lite logs

### LINKED ISSUE #455: Create a logging UI in KA Lite windows application
## Summary

> This will show all the logs when starting kalite.
> This will also fix #437

mrpau-richard:
## Summary

> I created a new menu to show the KA Lite logs using LogExpet application. 

## Issues addressed

Fixes #455 

@radinamatic Here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMcWl5OVRlcVN6VGM) to test this.


radinamatic:
@mrpau-richard I wasn't aware this required additional .NET framework install:

logexpet1

It wasn't problematic to install, I'm just not sure we should require it from users...

logexpet2

Would this work on older versions of Windows? How about if they can't connect the device to Internet to download the .NET framework installer? 

Could we try to implement the option for viewing logs without additional installation? What's wrong with plain ol' Notepad? They had that back on Win 95... 😉 

@benjaoming Thoughts?

benjaoming:
I agree with @radinamatic - we should require external software for a trivial thing such as logs. The most important part is that users are aware that they exist and can easily open them and copy-paste stuff if they need to report about bugs etc.

Also, in case of crashes, it seems like a good idea to open up the log in Notepad so it won't disappear. In case of `server.log` it's not so dangerous in terms of file size because the log is truncated every time KA Lite is started.

mrpau-richard:
@radinamatic thanks for testing this, I've removed the external application that's required .net application and used notepad as you suggested. here's the [installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMbzRsX3RaQzRGdU0) for another testing.

radinamatic:
@mrpau-richard Perfect, `Notepad` is back! 👍 

virtualbox_msedge - win10_preview_12_06_2017_14_22_24

benjaoming:
@mrpau-richard do you need a review from @cpauya? In case not, I recommend self-merging.

windows/README.md
```diff
@@ -8,6 +8,7 @@ This project provides a smoother way to install and run KA Lite in a Windows Mac
 * Inno Setup 5.5.3 [Download] (http://files.jrsoftware.org/is/5/)
 * Microsoft Visual Studio Community 2015 [Website] (https://www.visualstudio.com/)
 * Git (note: install with the option to place the `git` executable in the path, so it can be run within `cmd`)
+* LogExpert Website] (http://www.log-expert.de)
```
cpauya: We no longer use this.

windows/gui-source/KA Lite/KALite.cpp
```diff
@@ -209,6 +211,25 @@ void autoStartServerAction()
...
 
+void showKaliteServerLogs()
+{	
+	const DWORD MAX_SIZE = 255;
+	std:string filePath = "\\server.log";
+	char home_path[MAX_SIZE];
+	kaliteHomePath(home_path, MAX_SIZE);
+	std::string str(home_path);
+	struct stat fileAtt;
+	std::string kalite_logs_path = home_path + filePath;
```
cpauya: Inconsistent variable naming - this is C++ and not Python - your variables should use camelCase.  Here's an MSDN reference that you can apply: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions.

windows/gui-source/KA Lite/KALite.cpp
```diff
@@ -282,6 +303,7 @@ int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE hPrevInstance, LPSTR lpCmdLine
 	menu6 = new fle_TrayMenuItem("Run KA Lite at system startup.", &runAtStartupAction);
```
cpauya: Why are we still using ambiguous variable names like `menu6` here instead of `mnuRunAtStartup` or similar?  

I remember having reviewed a PR with these variables which you've already corrected @mrpau-richard - or maybe it's on Kolibri?  If so, then we should propagate those good programming practices back here at KA-Lite.

You can create a separate PR for these changes if you want.

mrpau-richard: Ok, I'll create another PR for that one.

mrpau-richard:
@cpauya I'm done fixing all your comments.

cpauya:
Nice one @mrpau-richard - let's test it one more time @mrpau-eduard then you can merge away.

radinamatic:
Ready to test again, if necessary! 👍

mrpau-eduard:
Hi @radinamatic you can grab the [windows installer](https://drive.google.com/open?id=0B-E59G3J_sXmN1E0WDdld1lRRDQ) here. I built a new one for this feature so you can test it. 
So far it is working on our side.

mrpau-eduard:
Done testing this feature with @mrpau-eugene .
Merging this `PR` so we can use it :)

PR #462, opened 2017-07-21
Allow user to import KA Lite database during upgrade.

### LINKED ISSUE #459: Allow user to import KA Lite database during upgrade process
## Summary

> This issue came up after allowing the user to use their KA Lite content path in this [PR](https://github.com/learningequality/ka-lite-installers/pull/442)

mrpau-richard:
## Summary

> I created a dialog message that allow the user to import KA Lite database during upgrade.

## Issues addressed

Fixes #459 

Here's the [windows installer](https://drive.google.com/open?id=0B5Tyi-tRKOFMalVyNTFEN010OHc) for testing.

/cc @radinamatic  @benjaoming

### DIFF
--- a/windows/installer-source/KaliteSetupScript.iss
+++ b/windows/installer-source/KaliteSetupScript.iss
@@ -289,7 +289,6 @@ var
     prevVerStr : String;
     retCode: Integer;
 begin
-    SelectKaliteDatabase;
     prevVerStr := GetPreviousVersion();
     if (CompareStr('{#TargetVersion}', prevVerStr) >= 0) and not (prevVerStr = '') then
     begin
@@ -305,8 +304,10 @@ begin
         end
         else
         begin
-            { This is where version-specific migration stuff should happen. }
 
+            SelectKaliteDatabase;
+
+            { This is where version-specific migration stuff should happen. }
             if CompareStr(prevVerStr, '0.13.99') < 0 then
             begin
                 if CompareStr('{#TargetVersion}', '0.14.0') >= 0 then
@@ -359,6 +360,7 @@ begin
         { Unclear what the logic here is. This is only executed if HandleUpgrade was not previously run. }
         if Not isUpgrade then
             HandleUpgrade(ExpandConstant('{app}'));
+
     end;
 end;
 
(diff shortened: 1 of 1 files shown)


benjaoming:
@mrpau-richard sorry that there have been no reviews here

Could you explain why importing a database is only relevant during an upgrade?

mrpau-richard:
@benjaoming I think the reason why I have this PR is when we allow the user to set there own KA Lite content path, it may also need to import there existing KA Lite database. I suggested it in this [PR](https://github.com/learningequality/ka-lite-installers/pull/442#issuecomment-303934115).

mrpau-richard:
@benjaoming it's also applicable to import database during fresh installation. Sorry I did not get your question [here](https://github.com/learningequality/ka-lite-installers/pull/462#issuecomment-326426522).

benjaoming:
@mrpau-richard would you say that this is safe to merge?

If an upgrade through the Windows Installer causes the database to be overwritten presently, then it would sound like a critical fix.

However, if the current behavior also keeps an existing database unharmed, then maybe we should skip this PR, seeing that we probably won't be able to maintain any possible bugs that it might cause (not saying that the code is bad, just that it's rather complex by nature) ?

mrpau-richard:
@benjaoming I just re-tested this, During the upgrade it will overwrite the existing KA Lite database and replace with the new one that the user uploaded.

benjaoming:
@mrpau-richard wasn't the old behavior during upgrade to keep the user's existing database?

I mean, the same as what would happen if you `pip install ka-lite` (first time) and then when an upgrade arrives `pip install ka-lite --upgrade`: You just get a newer version, but everything in `~/.kalite` remains the same...

mrpau-richard:
@benjaoming Yeah, During upgrade the old KA Lite database remain. Unless the user uploaded a new one and it replace the existing database.

benjaoming:
Thanks for the clarification @mrpau-richard -- then I would say that this issue is not important: **Replacing** the database while upgrading does not fit any use case that I know of. **Preserving** the database is the critical part, and if that's already covered, then I think it's okay to move on.

Would you be okay about closing this PR? We have a lot of stuff going on :)

mrpau-richard:
I agree @benjaoming lets close this and move on.

PR #463, opened 2017-07-27
Fixing `service ka-lite status` command

benjaoming:
## Summary

thanks @jamalex!

## TODO

- [x] Updated installer's local changelog
- [ ] Verified test builds


benjaoming:
@jamalex - should work fine both for systemd and sysVinit

image

